PR #7, opened 2023-04-17
fix: Teach update about split apks

eseidel:
Fixes https://github.com/shorebirdtech/shorebird/issues/292.

The previous code only worked with universal apks, not split apks (which is what the play store produces).

We now look for APKs in the same directory as base.apk (through the same path hacks as before) and look for one matching the arch/abi name first.  If we don't find one we also check base.apk.

This works in my testing with bundletool locally.

I also removed support for hashes being optional and removed a few unwraps (which will crash on failure).

I also fixed our use of imports/use statements to be a bit more consistent, including using "std::fs" rather than "std::fs::File" in one place but not another.


felangel (approved):
🎉

felangel:
Looks like a test failed: `Odd number of digits', library/src/updater.rs:552:61`

eseidel:
> Looks like a test failed: `Odd number of digits', library/src/updater.rs:552:61`

Fixed!

PR #11, opened 2023-04-24
feat: Add start_updater_thread to update off the main thread

eseidel:
This makes it so that clients can easily not block when wanting to
queue an update.

I have a separate patch which updates the Engine to use this new
API.

I also needed to split the concept of the "next_boot" patch
from the "current_boot" patch, previously refered to as
"current" or "active" patch.  This required adding a
report_launch_start api to let the updater library know
when to set current_boot patch from next_boot.

I also removed the rust updater/cli in this as well as the
vmpath argument to init.

I also exposed the report_launch_success api, but its not yet
used by the Engine.

I renamed report_failed_launch to report_launch_failure to match
report_launch_start which I introduced.


library/src/c_api.rs
```diff
@@ -96,9 +89,8 @@ pub extern "C" fn shorebird_active_patch_number() -> *mut c_char {
 /// Return the path to the active patch for the app, or NULL if there is no
 /// active patch.
 #[no_mangle]
-// rename to shorebird_patch_path
-pub extern "C" fn shorebird_active_path() -> *mut c_char {
-    let version = updater::active_patch();
+pub extern "C" fn shorebird_next_boot_patch_path() -> *mut c_char {
+    let version = updater::next_boot_patch();
```
felangel, line 93:
```suggestion
    let patch_number = updater::next_boot_patch();
```

nit but version might not be the most descriptive name for this 🤷

library/src/updater.rs
```diff
@@ -437,6 +456,19 @@ pub fn update() -> UpdateStatus {
     });
 }
 
+/// This does not return status.  The only output is the change to the saved
+/// cache.  The Engine calls this during boot and it will check for an update
```
felangel, line 460:
```suggestion
/// cache. The Engine calls this during boot and it will check for an update
```

library/src/updater.rs
```diff
@@ -437,6 +456,19 @@ pub fn update() -> UpdateStatus {
...
+/// and install it if available.
+pub fn start_update_thread() {
+    // This holds the lock on the config for the entire duration of the update
```
felangel, line 463: Is there any harm in holding the lock on the config?

felangel (approved):

library/src/updater.rs
eseidel: There is, in that any other calls into the library will block until the lock is released. Which means that if we add a dart api those calls would hang while an update is in progress.  Better is to have it drop the lock while it's doing network requests since those can go on for a long time.

PR #14, opened 2023-04-25
test: Make our rust tests use per-thread config rather than global static

eseidel:
This lets our unit tests run in parallel.  I also moved our "integration"
tests back to be unit tests for now.

This also exposes a network mocking abstraction.

I've added an incomplete test of applying a full patch.  Still
needs some work to complete.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


library/src/config.rs
```diff
@@ -13,27 +20,127 @@ const DEFAULT_BASE_URL: &'static str = "https://api.shorebird.dev";
...
+        Self {
+            resolved_config: ResolvedConfig::empty(),
+            patch_check_request_fn: None,
+            download_file_fn: None,
+        }
+    }
+}
+
+#[cfg(test)]
+// Used only for testing.  Could be guarded by the test feature flag?
```
eseidel, line 48: lol.  Can be removed.

library/src/config.rs
```diff
@@ -13,27 +20,127 @@ const DEFAULT_BASE_URL: &'static str = "https://api.shorebird.dev";
...
+thread_local!(static THREAD_CONFIG: RefCell<ThreadConfig> = RefCell::new(ThreadConfig::empty()));
+
+#[cfg(test)]
+/// This is used by tests to change how we store config to be per-thread.
```
eseidel, line 52: Also wrong now.

library/src/config.rs
```diff
@@ -13,27 +20,127 @@ const DEFAULT_BASE_URL: &'static str = "https://api.shorebird.dev";
...
-    }
-    return f(&lock);
+#[cfg(test)]
+pub fn with_config<F, R>(f: F) -> R
+where
+    F: FnOnce(&ResolvedConfig) -> R,
+{
+    // Rust unit tests run on multiple threads in parallel, so we use
+    // a per-thread config when unit testing instead of a global config.
+    // The global config code paths are covered by the integration tests.
```
eseidel, line 112: I should probably add one back before I check this in.

eseidel: Will do that later.  This patch is too big.

library/src/config.rs
eseidel: Done.

library/include/updater.h
```diff
@@ -48,10 +48,11 @@ extern "C" {
 /**
  * Configures updater.  First parameter is a struct containing configuration
  * from the running app.  Second parameter is a YAML string containing
- * configuration compiled into the app.
+ * configuration compiled into the app.  Returns true on success and false on
+ * failure. If false is returned, the updater library will not be usable.
```
felangel, line 52: Should we include what should be done in the case it returns false? Would re-running `shorebird_init` be advised? Is there any way to recover?

felangel (approved):
LGTM

PR #18, opened 2023-05-16
fix: don't panic on missing shorebird.yaml

bryanoltman:
## Description

Fixes https://github.com/shorebirdtech/shorebird/issues/413

Validated by:
- Creating a new Flutter project
- Running `shorebird init`
- Removing `shorebird.yaml`
- Verifying that `shorebird --local-engine-src-path=$HOME/Shorebird/engine/src --local-engine=android_release_arm64 run -- --release` doesn't panic and still launches the app.

I also tested the release-patch flow to ensure that didn't break.

- Updates `with_config` in `library/src/config.rs` to
  - Return `anyhow::Result<R>` instead of `R`
  - Not panic if the config object isn't initialized
  - Return an error in the case where the config object isn't initialized
- Updates consumers of `with_config` to handle the possibility that `with_config` may return an error by propagating the error, usually to `log_on_error` in `library/src/c_api.rs` because the updater can't do much if no `shorebird.yaml` is provided.

## Potential Future Work

- Improve use of `anyhow::Error`
  - Wrapping our error types in `anyhow::Error::from()` is pretty verbose and I suspect there's a better, more idiomatic way to use `anyhow::Error`. The `report_launch_result_with_no_current_patch` test is especially ugly.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/c_api.rs
+++ b/library/src/c_api.rs
@@ -83,14 +83,10 @@ fn log_on_error<F, R>(f: F, context: &str, error_result: R) -> R
 where
     F: FnOnce() -> Result<R, anyhow::Error>,
 {
-    let result = f();
-    match result {
-        Ok(r) => r,
-        Err(e) => {
-            error!("Error {}: {:?}", context, e);
-            error_result
-        }
-    }
+    f().unwrap_or_else(|e| {
+        error!("Error {}: {:?}", context, e);
+        error_result
+    })
 }
 
 /// Configures updater.  First parameter is a struct containing configuration
@@ -119,7 +115,7 @@ pub extern "C" fn shorebird_init(
 pub extern "C" fn shorebird_next_boot_patch_number() -> *mut c_char {
     log_on_error(
         || {
-            let patch = updater::next_boot_patch().context("failed to fetch patch info")?;
+            let patch = updater::next_boot_patch()?.context("failed to fetch patch info")?;
             allocate_c_string(&patch.number.to_string())
         },
         "fetching next_boot_patch_number",
@@ -133,7 +129,7 @@ pub extern "C" fn shorebird_next_boot_patch_number() -> *mut c_char {
 pub extern "C" fn shorebird_next_boot_patch_path() -> *mut c_char {
     log_on_error(
         || {
-            let patch = updater::next_boot_patch().context("failed to fetch patch info")?;
+            let patch = updater::next_boot_patch()?.context("failed to fetch patch info")?;
             allocate_c_string(&patch.path)
         },
         "fetching next_boot_patch_path",
@@ -155,13 +151,20 @@ pub extern "C" fn shorebird_free_string(c_string: *mut c_char) {
 /// Check for an update.  Returns true if an update is available.
 #[no_mangle]
 pub extern "C" fn shorebird_check_for_update() -> bool {
-    return updater::check_for_update();
+    log_on_error(updater::check_for_update, "checking for update", false)
 }
 
 /// Synchronously download an update if one is available.
 #[no_mangle]
 pub extern "C" fn shorebird_update() {
-    updater::update();
+    log_on_error(
+        || {
+            updater::update()?;
+            Ok(())
+        },
+        "downloading update",
+        (),
+    );
 }
 
 /// Start a thread to download an update if one is available.
@@ -177,11 +180,7 @@ pub extern "C" fn shorebird_start_update_thread() {
 /// shorebird_report_launch_success or shorebird_report_launch_failure.
 #[no_mangle]
 pub extern "C" fn shorebird_report_launch_start() {
-    log_on_error(
-        || updater::report_launch_start(),
-        "reporting launch start",
-        (),
-    );
+    log_on_error(updater::report_launch_start, "reporting launch start", ());
 }
 
 /// Report that the app failed to launch.  This will cause the updater to
@@ -190,10 +189,7 @@ pub extern "C" fn shorebird_report_launch_start() {
 #[no_mangle]
 pub extern "C" fn shorebird_report_launch_failure() {
     log_on_error(
-        || {
-            updater::report_launch_failure()?;
-            Ok(())
-        },
+        updater::report_launch_failure,
         "reporting launch failure",
         (),
     );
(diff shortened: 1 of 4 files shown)


library/src/c_api.rs
```diff
@@ -119,7 +115,7 @@ pub extern "C" fn shorebird_init(
 pub extern "C" fn shorebird_next_boot_patch_number() -> *mut c_char {
     log_on_error(
         || {
-            let patch = updater::next_boot_patch().context("failed to fetch patch info")?;
+            let patch = updater::next_boot_patch()?.context("failed to fetch patch info")?;
```
eseidel, line 118: Is this now a Result<Result<T>>?  .context is a method on result (from anyhow) to add context to the potential error.

library/src/config.rs
```diff
@@ -60,14 +62,14 @@ pub fn testing_reset_config() {
     });
 }
 
-pub fn check_initialized_and_call<F, R>(f: F, config: &ResolvedConfig) -> R
+pub fn check_initialized_and_call<F, R>(f: F, config: &ResolvedConfig) -> anyhow::Result<R>
 where
     F: FnOnce(&ResolvedConfig) -> R,
```
eseidel, line 67: To avoid the Result<Result<T>> we can specify that F returns a Result no?

library/src/updater.rs
```diff
@@ -89,26 +91,24 @@ pub fn init(app_config: AppConfig, yaml: &str) -> Result<(), UpdateError> {
...
-pub fn check_for_update() -> bool {
-    return with_config(check_for_update_internal);
+pub fn check_for_update() -> anyhow::Result<bool> {
+    with_config(|config| {
+        // Load UpdaterState from disk
+        // If there is no state, make an empty state.
+        let state = UpdaterState::load_or_new_on_error(&config.cache_dir, &config.release_version);
+        // Send info from app + current slot to server.
+        let response_result = send_patch_check_request(&config, &state);
+        match response_result {
```
eseidel, line 102: I think this can just return the result instead of logging about it here.

library/src/updater.rs
```diff
@@ -421,41 +421,51 @@ pub fn report_launch_start() -> anyhow::Result<()> {
...
         state.activate_current_patch()?;
         state.save()
-    })
+    })?
 }
 
 /// Report that the current active path failed to launch.
 /// This will mark the patch as bad and activate the next best patch.
-pub fn report_launch_failure() -> Result<(), UpdateError> {
+pub fn report_launch_failure() -> anyhow::Result<()> {
```
eseidel, line 429: UpdateError was a dream.  Probably can die now that we've teched up to anyhow.

eseidel (approved):
let chat on vc quick.

library/src/c_api.rs
eseidel: I think this is because it's now Result<Option>?

eseidel: This calls ? to unwrap the Result<Option<T>> into Option<T> which then context() turns back into a Result<T> which we then unwrap again. 🤣   We should just be able to return the option w/o logging?

library/src/config.rs
bryanoltman: If I understand correctly, updating the constraints on `F` to return `Result<R>` instead of `R` will take us in the wrong direction (it will add an extra layer of `Result`). 

`check_initialized_and_call` needs to return a `Result`, possibly without invoking `F`, because it bails if `config` is not initialized. `R` _can_ be a `Result` (and sometimes is), but the code is usually cleaner if it isn't (see `check_for_update`, which would need to return `Ok(false)` instead of just `false` if an update is not available).

library/src/updater.rs
bryanoltman: Filed https://github.com/shorebirdtech/updater/issues/19 – I think we'll want to overhaul our error handling more generally as part of this.

library/src/c_api.rs
```diff
@@ -155,13 +157,20 @@ pub extern "C" fn shorebird_free_string(c_string: *mut c_char) {
...
+    log_on_error(updater::check_for_update, "checking for update", false)
 }
 
 /// Synchronously download an update if one is available.
 #[no_mangle]
 pub extern "C" fn shorebird_update() {
-    updater::update();
+    log_on_error(
+        || {
+            updater::update()?;
```
eseidel, line 168: This closure isn't necessary.

library/src/config.rs
```diff
@@ -1,4 +1,5 @@
 // This file handles the global config for the updater library.
+use anyhow::bail;
```
eseidel, line 2: I don't bother to import things when I'm only calling them once. 🤷

library/src/c_api.rs
eseidel: Actually it is, but another way would be to catch the update status, match it and log from the match.  (Or just log the update status and assume it has a Display property.

library/src/updater.rs
```diff
@@ -89,26 +91,24 @@ pub fn init(app_config: AppConfig, yaml: &str) -> Result<(), UpdateError> {
...
-    return with_config(check_for_update_internal);
+pub fn check_for_update() -> anyhow::Result<bool> {
+    with_config(|config| {
+        // Load UpdaterState from disk
+        // If there is no state, make an empty state.
+        let state = UpdaterState::load_or_new_on_error(&config.cache_dir, &config.release_version);
+        // Send info from app + current slot to server.
+        let response_result = send_patch_check_request(&config, &state);
+        match response_result {
+            Err(err) => {
```
eseidel: In a follow up patch, we should just make the Rust API return these errors, and the c_api responsible to log them when needed.

library/src/config.rs
bryanoltman: Updated as per offline convo

library/src/updater.rs
bryanoltman: Done.

library/src/config.rs
bryanoltman: Fixed.

library/src/c_api.rs
bryanoltman: Fixed.

eseidel (approved):

PR #23, opened 2023-06-08
feat: Support iOS

eseidel:
Adds support for iOS to the updater library.

Refactored the android-specific code into an `android` module.

The tests still mostly use the `android` code, rather than testing iOS paths too, but that can be fixed in a follow-up.

I had to change how "latest patch number" works to understand when patch files disappear from disk (which I thought was happening on iOS initially).

I also had to make status.json not store absolute paths to patch files (since the absolute path changes per-xcode run on iOS).


library/src/cache.rs
```diff
@@ -179,13 +182,44 @@ impl UpdaterState {
...
+        Ok(())
+    }
+
     fn validate_slot(&self, slot: &Slot) -> bool {
         // Check if the patch is known bad.
-        if self.is_known_bad_patch(&slot.to_patch_info()) {
+        if self.is_known_bad_patch(&PatchInfo {
+            path: String::new(),
+            number: slot.patch_number,
+        }) {
```
bryanoltman, line 211: I'd add a fixme here to update is_known_bad_patch to only require a patch number

library/src/updater.rs
```diff
@@ -312,8 +162,39 @@ fn update_internal(config: &ResolvedConfig) -> anyhow::Result<UpdateStatus> {
     debug!("app_dir: {:?}", app_dir);
     let base_r = open_base_lib(&app_dir, "libapp.so")?;
 
+    inflate(&download_path, base_r, &output_path)
+}
+
+#[cfg(not(any(target_os = "android", test)))]
```
bryanoltman, line 168: Thoughts on making this explicit that we're targeting iOS (as opposed to "not Android"). Not sure how doable that is since we're also making sure this isn't a test cfg

bryanoltman (approved):

library/src/cache.rs
```diff
@@ -154,7 +154,10 @@ impl UpdaterState {
             return None;
         }
         let slot = &self.slots[index];
-        Some(slot.to_patch_info())
+        Some(PatchInfo {
+            path: self.patch_path(index).to_str().unwrap().to_owned(),
```
felangel, line 158: Is it safe to unwrap here?

library/src/updater.rs
eseidel: In this case iOS *should* use this function, just doesn't yet.  That's why iOS patch files are MB instead of bytes.

library/src/cache.rs
```diff
@@ -179,13 +182,44 @@ impl UpdaterState {
...
+        if needs_save {
+            self.save()?;
+        }
+        Ok(())
+    }
+
     fn validate_slot(&self, slot: &Slot) -> bool {
         // Check if the patch is known bad.
-        if self.is_known_bad_patch(&slot.to_patch_info()) {
+        if self.is_known_bad_patch(&PatchInfo {
```
felangel, line 208: Can we change the signature to only accept a patch number here so we don’t need to pass an empty path?

library/src/cache.rs
```diff
@@ -179,13 +182,44 @@ impl UpdaterState {
...
+            info!("Slot {:?} is known bad.", slot);
             return false;
         }
-        if PathBuf::from(&slot.path).exists() {
-            return true;
+        let index = self
+            .slots
+            .iter()
+            .position(|s| s.patch_number == slot.patch_number);
+        let patch_path = self.patch_path(index.unwrap());
```
felangel, line 219: Same here about unwrapping and whether it’s safe?

felangel (approved):

PR #24, opened 2023-06-08
fix: remove more unwrap calls.

eseidel:
Each unwrap() is a potential crash.  Shorebird's updater code should never crash.

There still are a couple, but I think we should set up coverage first so we can track that we're testing all these changes.


felangel (approved):
💯

PR #32, opened 2023-06-19
feat(shorebird_code_push): add updater bindings and dart support

bryanoltman:
## Description

Adds ffi bindings for the Updater and support for using them.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


shorebird_code_push/README.md
```diff
@@ -4,6 +4,38 @@
 
 A Dart library that allows Flutter apps to get information about code push updates.
 
+## Usage
+
+```dart
+import 'package:shorebird_code_push/shorebird_code_push.dart';
+
+final shorebirdCodePush = ShorebirdCodePush();
```
eseidel, line 12: I wonder if we should drop our name from the classes in the package?

shorebird_code_push/README.md
```diff
@@ -4,6 +4,38 @@
...
+The Dart code in this library communicates with the Updater (part of Shorebird's
+Flutter engine) via FFI.
+
+For an Updater function to be visible to the Dart code, it must:
+
+1. Be declared in c_api.rs as `pub extern "C"`.
+    1. This will add the function to the `library/include/updater.h` header
+    file, which is generated by [cbindgen](https://github.com/mozilla/cbindgen)
+    when the Updater is built.
+1. Be listed in
```
eseidel, line 34: This is only for Android (iOS will have a different file I suspect).

shorebird_code_push/lib/src/generated/updater_bindings.g.dart
```diff
@@ -0,0 +1,4421 @@
...
+
+const int BYTE_ORDER = 1234;
+
+const int NULL = 0;
+
+const int EXIT_FAILURE = 1;
+
+const int EXIT_SUCCESS = 0;
+
+const int RAND_MAX = 2147483647;
```
eseidel, line 4421: Wow.  I feel like I wired these up by hand in about 20 lines of code.  Not objecting to using a generator, just surprised at how much gunk it spits out.

shorebird_code_push/lib/src/updater.dart
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /// Creates an [Updater] instance using the dynamic library at the given
+  /// [directory].
+  ///
+  /// The name of the library is determined by the current OS:
+  ///   macOS:   lib[name].dylib
+  ///   Windows: [name].dll
+  ///   Linux:   lib[name].so
+  Updater.initWithLibrary({
```
eseidel, line 26: This is only needed in a pure Dart environment (which we doubt we support anymore unless you've resurrected it).

shorebird_code_push/lib/src/updater.dart
```diff
@@ -0,0 +1,114 @@
...
+    bindings = UpdaterBindings(dylib);
+  }
+
+  /// The ffi bindings to the Updater library.
+  @visibleForTesting
+  static late UpdaterBindings bindings;
+
+  /// An override for dart:io's [Platform] class for testing.
+  @visibleForTesting
+  static Platform platform = const LocalPlatform();
```
eseidel, line 51: Isn't this package included in customer code?  I suspect we want to minimize dependencies as much as possible. 🤷

shorebird_code_push/lib/src/updater.dart
```diff
@@ -0,0 +1,114 @@
...
+    );
+    return patchNumberString == null ? null : int.tryParse(patchNumberString);
+  }
+
+  /// Whether a new patch is available.
+  bool checkForUpdate() => bindings.shorebird_check_for_update();
+
+  /// A wrapper for ffi functions that return [Pointer<Char].
+  String? _returnsMaybeString(ffi.Pointer<ffi.Char> Function() f) {
+    var cString = ffi.Pointer<ffi.Char>.fromAddress(0);
```
eseidel, line 68: Why does this need to be initialized separately first?

eseidel: I could imagine it would want to be if its setter is done inside a try, but it doesn't seem to be the case here?

shorebird_code_push/pubspec.yaml
```diff
@@ -5,9 +5,21 @@ version: 0.1.0+1
 environment:
   sdk: ">=3.0.0 <4.0.0"
 
+dependencies:
```
eseidel, line 8: I suspect we'll need/want to reduce this set of dependencies to avoid introducing conflicts for customers.  We don't have to do that now, but likely want to leave a fixme. 🤷

eseidel (approved):
It's a good start. Please consider my comments.

shorebird_code_push/lib/src/updater.dart
bryanoltman: I'll remove for now. It'll always be in commit history if we want to resurrect.

shorebird_code_push/lib/src/generated/updater_bindings.g.dart
bryanoltman: Yeah, it is a lot. Didn't have to write a line, but there's probably some way to cut back on the amount being generated.

shorebird_code_push/lib/src/updater.dart
bryanoltman: Removed, along with other vestigial Dart CLI code

shorebird_code_push/lib/src/updater.dart
bryanoltman: ¯\_(ツ)_/¯ this was pulled from the old dart bindings code. Removed.

shorebird_code_push/pubspec.yaml
bryanoltman: Removed, along with other vestigial Dart CLI code

shorebird_code_push/README.md
bryanoltman: Addressed

shorebird_code_push/README.md
bryanoltman: No strong opinion here. I'd say we just leave it as is for now – it's easy enough to change later

eseidel: It just requires all clients to update, but that's fine too.  We can manage that when we do so.

PR #33, opened 2023-06-19
feat(shorebird_code_push): update example to use code push bindings

bryanoltman:
## Description

Updates the `shorebird_code_push` example app to use the updater bindings.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


shorebird_code_push/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -1,33 +1,19 @@
...
-        <activity
-            android:name=".MainActivity"
-            android:exported="true"
-            android:launchMode="singleTop"
-            android:theme="@style/LaunchTheme"
-            android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
-            android:hardwareAccelerated="true"
-            android:windowSoftInputMode="adjustResize">
-            <!-- Specifies an Android theme to apply to this Activity as soon as
+  <application android:label="example" android:name="${applicationName}" android:icon="@mipmap/ic_launcher">
```
eseidel, line 2: Is this just reformatting?

shorebird_code_push/example/lib/main.dart
```diff
@@ -1,125 +1,82 @@
...
+  void initState() {
+    super.initState();
     setState(() {
-      // This call to setState tells the Flutter framework that something has
-      // changed in this State, which causes it to rerun the build method below
-      // so that the display can reflect the updated values. If we changed
-      // _counter without calling setState(), then the build method would not be
-      // called again, and so nothing would appear to happen.
-      _counter++;
+      _patchVersion = shorebirdCodePush.currentPatchVersion();
```
eseidel, line 44: This can hang indefinitely if the lock is currently held by something else (like the background thread pulling an update). I like your suggestion of making it async.

shorebird_code_push/example/lib/main.dart
```diff
@@ -1,125 +1,82 @@
...
     });
   }
 
+  Future<void> _checkForUpdate() async {
+    _isUpdateAvailable = await shorebirdCodePush.checkForUpdate();
```
eseidel, line 49: Shouldn't this line technically move inside the await?  Or I guess technically the await is supposed to return to a local and then teh setting from the local is inside setState?  I'm not sure what ideal dart patterns are.

eseidel (approved):
I think this is a good step.  You'll want more review from felix (or someone else with more sense of what "great flutter" looks like -- since I don't really have it) as we get further along I suspect.

shorebird_code_push/example/lib/main.dart
bryanoltman: Done.

shorebird_code_push/example/lib/main.dart
bryanoltman: Updated

shorebird_code_push/lib/src/generated/updater_bindings.g.dart
```diff
@@ -2084,20 +2084,20 @@ class UpdaterBindings {
...
-  ffi.Pointer<ffi.Char> shorebird_next_boot_patch_number() {
+  /// The patch number that will boot on the next run of the app, or 0 if there is
+  /// no next patch.
+  int shorebird_next_boot_patch_number() {
     return _shorebird_next_boot_patch_number();
   }
 
   late final _shorebird_next_boot_patch_numberPtr =
-      _lookup<ffi.NativeFunction<ffi.Pointer<ffi.Char> Function()>>(
+      _lookup<ffi.NativeFunction<ffi.UintPtr Function()>>(
```
eseidel, line 2094: FWIW, We will end up needing to handle both partial and full failure of lookups in this code.  partial will happen when there is version skew between this package and the version of shorebird in use (wrong verison pulled from pub) and full failure will happen every time the user uses `flutter` directly instead of `shorebird`.

eseidel (approved):

shorebird_code_push/example/lib/main.dart
```diff
@@ -1,125 +1,108 @@
...
+      });
+    });
+  }
+
+  Future<void> _checkForUpdate() async {
+    setState(() {
+      _isCheckingForUpdate = true;
+    });
+
+    bool isUpdateAvailable = await shorebirdCodePush.checkForUpdate();
```
felangel, line 55:
```suggestion
    final isUpdateAvailable = await shorebirdCodePush.checkForUpdate();
```

shorebird_code_push/example/pubspec.yaml
```diff
@@ -1,90 +1,24 @@
...
-# build by specifying --build-name and --build-number, respectively.
-# In Android, build-name is used as versionName while build-number used as versionCode.
-# Read more about Android versioning at https://developer.android.com/studio/publish/versioning
-# In iOS, build-name is used as CFBundleShortVersionString while build-number is used as CFBundleVersion.
-# Read more about iOS versioning at
-# https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
-# In Windows, build-name is used as the major, minor, and patch parts
-# of the product and file versions while build-number is used as the build suffix.
-version: 1.0.0+1
+version: 1.0.0+15
```
felangel, line 5: why is this build 15?

shorebird_code_push/example/lib/main.dart
```diff
@@ -1,125 +1,108 @@
 import 'package:flutter/material.dart';
 
+import 'package:shorebird_code_push/shorebird_code_push.dart';
+
+final ShorebirdCodePush shorebirdCodePush = ShorebirdCodePush();
```
felangel, line 5: does this need to be a global variable? Could it be initialized as part of the `StatefulWidget` `State`?

shorebird_code_push/example/lib/main.dart
```diff
@@ -1,125 +1,108 @@
...
 class _MyHomePageState extends State<MyHomePage> {
-  int _counter = 0;
+  int? _patchVersion;
+  var _isCheckingForUpdate = false;
+
+  @override
+  void initState() {
+    super.initState();
+    shorebirdCodePush.currentPatchVersion().then((version) {
+      setState(() {
```
felangel, line 44: shouldn't we check if the widget is still mounted before calling `setState` here?

shorebird_code_push/example/lib/main.dart
```diff
@@ -1,125 +1,108 @@
...
 
-  void _incrementCounter() {
     setState(() {
-      // This call to setState tells the Flutter framework that something has
-      // changed in this State, which causes it to rerun the build method below
-      // so that the display can reflect the updated values. If we changed
-      // _counter without calling setState(), then the build method would not be
-      // called again, and so nothing would appear to happen.
-      _counter++;
+      _isCheckingForUpdate = false;
```
felangel, line 58: same here shouldn't we do the mounted check before calling `setState`?

shorebird_code_push/example/lib/main.dart
felangel, line 5:
```suggestion
final shorebirdCodePush = ShorebirdCodePush();
```

Would be nice to use the same analysis options we use in the shorebird repo for consistency.

shorebird_code_push/example/pubspec.yaml
bryanoltman: I've been using this to test and have released 1.0.0+1 to 1.0.0+15

shorebird_code_push/example/lib/main.dart
bryanoltman: Good call, done

shorebird_code_push/example/lib/main.dart
bryanoltman: Huh, not sure why the analyzer was OK with this and not the `ScaffoldMessenger.of(context)` below. Fixed!

PR #34, opened 2023-06-20
chore: use a usize instead of a char* for next boot patch number

bryanoltman:
## Description

Passes the next patch number back as a `usize` instead of a `char *`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/library/include/updater.h
+++ b/library/include/updater.h
@@ -56,13 +56,14 @@ bool shorebird_init(const struct AppParameters *c_params,
                     const char *c_yaml);
 
 /**
- * Return the active patch number, or NULL if there is no active patch.
+ * The patch number that will boot on the next run of the app, or 0 if there is
+ * no next patch.
  */
-SHOREBIRD_EXPORT char *shorebird_next_boot_patch_number(void);
+SHOREBIRD_EXPORT uintptr_t shorebird_next_boot_patch_number(void);
 
 /**
- * Return the path to the active patch for the app, or NULL if there is no
- * active patch.
+ * The path to the patch that will boot on the next run of the app, or NULL if
+ * there is no next patch.
  */
 SHOREBIRD_EXPORT char *shorebird_next_boot_patch_path(void);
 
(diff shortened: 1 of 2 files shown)


library/include/updater.h
```diff
@@ -56,13 +56,14 @@ bool shorebird_init(const struct AppParameters *c_params,
                     const char *c_yaml);
 
 /**
- * Return the active patch number, or NULL if there is no active patch.
+ * The patch number that will boot on the next run of the app, or 0 if there is
+ * no next patch.
  */
-SHOREBIRD_EXPORT char *shorebird_next_boot_patch_number(void);
+SHOREBIRD_EXPORT uintptr_t shorebird_next_boot_patch_number(void);
```
eseidel, line 62: You want usize here, I think?

eseidel (approved):
Other than the type in the header, this looks correct.  Please correct and land. :)

library/include/updater.h
bryanoltman: This is a C header, so I'm not sure `usize` exists. `usize` is "The pointer-sized unsigned integer type", so I think `uintptr_t` is the correct C-land representation of the bits.

eseidel: Oh yeah, sorry.  I *think* that's size_t then?  Which is in stddef.h?

eseidel: My data is https://locka99.gitbooks.io/a-guide-to-porting-c-to-rust/content/features_of_rust/types.html 🤷

bryanoltman: I saw that and https://stackoverflow.com/a/1464559 as well. fwiw, this header is autogenerated

eseidel: https://stackoverflow.com/questions/1464174/size-t-vs-uintptr-t

This stuff is all terrible. :)  Hence why we just write in higher level languages.

eseidel: > I saw that and https://stackoverflow.com/a/1464559 as well. fwiw, this header is autogenerated by cbindgen.

Oh.  Fine then.  I definitely trust the autogen more than my googling!  Thank you!

PR #44, opened 2023-06-22
fix: Only allow one update attempt at a time

eseidel:
Currently any shorebird_* call will hang if another update attempt is in progress.  Instead:
1. update should only hold the lock when absolutely necessary
2. Further attempts at updating should fail while an update is in progress.

This is all for https://github.com/shorebirdtech/shorebird/issues/695

I ended up fixing a lot of things in our Rust code while I was in there, including:
* Using PathBuf instead of Sting wherever we are holding onto Path objects (this removed a bunch of code which was converting in and out of these).
* Made a variety of functions private (the default) that were previously mistakenly `pub`.
* Moved the android hacks around libapp_paths closer to the C++ code.  Previously we were storing all of the libapp_paths on ResolvedConfig and handling the android hacks during patch install.  Now we're doing the android path hacks on init, eventually we'll remove them entirely by making the C++ code pass us in the android app_dir or even the resolved .apk we need instead of the libapp_paths.
* Fixed several tests to use /dir/lib/arch/libapp.so paths instead of 'libapp.so' now that the init code applies the android hacks an expects the deeper paths.  Again, this will go away when we move the android hacks into c++.
* Added a test which confirms that if we init twice we log instead of trying to init again (this was one behavior which FCM was triggering as part of  https://github.com/shorebirdtech/shorebird/issues/695
* Add a test which confirms that further calls to update while an update is ongoing just fail out quickly instead of hanging.
* Moved to serial_test instead of a separate per-thread config system for testing.  This makes our tests operate more like the code does in the wild, at the cost of needing to annotate a test as #[serial] any time it needs to use shorebird_init.
* Removed ResolvedConfig and now call it UpdateConfig.  ResolvedConfig previously had a bool on it to tell if it was initialized, that's now represented as Option<UpdateConfig> instead which is more rusty. 
* Changed how we handle NetworkHooks from being compile-time test switched to using function pointers held off of UpdateConfig.network_hooks.  I think this makes more sense?  I did this to remove the special ThreadConfig object which was used during unit tests, but didn't exist in production.  I think the way we mock networking could still be improved.
* There is now real risk of accidentally calling the real networking code in tests (which is what codecov is complaining about me not testing).  It's possible my NetworkHooks change is a bad idea?


library/src/android.rs
```diff
@@ -5,12 +5,17 @@ use std::path::{Path, PathBuf};
...
 // https://stackoverflow.com/questions/67087597/is-it-possible-to-use-rusts-log-info-for-tests
 #[cfg(test)]
-use std::println as debug; // Workaround to use println! for logs.
+use std::println as debug;
 
-pub(crate) fn app_data_dir_from_libapp_path(libapp_path: &str) -> anyhow::Result<PathBuf> {
+use crate::UpdateError; // Workaround to use println! for logs.
+
+fn app_data_dir_from_libapp_path(libapp_path: &str) -> Result<PathBuf, UpdateError> {
     // "/data/app/~~7LtReIkm5snW_oXeDoJ5TQ==/com.example.shorebird_test-rpkDZSLBRv2jWcc1gQpwdg==/lib/x86_64/libapp.so"
```
bryanoltman, line 13: For clarity, this is returning the `/data/app/~~7LtReIkm5snW_oXeDoJ5TQ==` directory?

library/src/network.rs
```diff
@@ -8,24 +8,44 @@ use std::path::Path;
...
 fn patches_check_url(base_url: &str) -> String {
     return format!("{}/api/v1/patches/check", base_url);
 }
 
-#[cfg(test)]
 pub type PatchCheckRequestFn = fn(&str, PatchCheckRequest) -> anyhow::Result<PatchCheckResponse>;
+pub type DownloadFileFn = fn(&str) -> anyhow::Result<Vec<u8>>;
+
+#[derive(Clone)]
+pub struct NetworkHooks {
```
bryanoltman, line 25: Documentation would be helpful here. "A container for network request callbacks, used for overriding default behavior in tests"

library/src/updater.rs
```diff
@@ -131,41 +163,20 @@ fn check_hash(path: &Path, expected_string: &str) -> anyhow::Result<bool> {
...
-        .last()
-        .context("No libapp paths")?;
-    // We could probably use sourceDir instead?
-    // https://developer.android.com/reference/android/content/pm/ApplicationInfo#sourceDir
-    // and splitSourceDirs (api 21+)
-    // https://developer.android.com/reference/android/content/pm/ApplicationInfo#splitSourceDirs
-    debug!("Finding apk from: {:?}", full_libapp_path);
-    let app_dir = app_data_dir_from_libapp_path(full_libapp_path)?;
+    // We abuse libapp_path to actually be the path to the data dir for now.
+    let app_dir = &config.libapp_path;
```
bryanoltman, line 171: Updating the comment to explain why this is abuse and what the correct behavior would be would be helpful

library/src/updater_lock.rs
```diff
@@ -0,0 +1,54 @@
+use crate::updater::UpdateError;
+
+// This file's job is to handle the boiler plate around locking for the
+// updater thread.
+// We should be able to share more of this code with similar boilerplate around
+// the ResolveConfig global lock with a little refactoring.
```
bryanoltman, line 6: Update to remove references to the deleted `ResolvedConfig`

library/src/updater.rs
```diff
@@ -174,8 +185,32 @@ fn prepare_for_install(
...
+    // Update will need to take the Config lock at times, but will only
+    // do so as long as is needed to read from the config and will not
+    // hold the config lock across network requests.
+    // Steps:
+    // Checks Update lock, if held, return error, otherwise takes lock for
+    // entire duration of update.
+    // Loads state from disk (holds Config lock while reading).
+    // Uses current update information to build request and send to server.
+    // If update is not available, returns.
+    // Update is avaiable, so uses returned information to download update.
```
felangel, line 206:
```suggestion
    // Update is available, so uses returned information to download update.
```

library/src/updater_lock.rs
```diff
@@ -0,0 +1,54 @@
+use crate::updater::UpdateError;
+
+// This file's job is to handle the boiler plate around locking for the
```
felangel, line 3:
```suggestion
// This file's job is to handle the boilerplate around locking for the
```

bryanoltman, line 3: I'd add a note about why we need to lock, e.g. "only one thread can mutate the file system at a time"

bryanoltman (approved):

library/src/c_api.rs
```diff
@@ -296,13 +299,15 @@ mod test {
         )
     }
 
+    #[serial]
```
felangel, line 302: Minor but it would be nice to create a module for tests as is described in the docs:
https://doc.rust-lang.org/book/ch11-03-test-organization.html

```rs
#[cfg(test)]
mod tests {
    #[test]
    fn it_works() {
        let result = 2 + 2;
        assert_eq!(result, 4);
    }
}
```

library/src/config.rs
```diff
@@ -25,78 +20,35 @@ const DEFAULT_BASE_URL: &'static str = "https://api.shorebird.dev";
...
+) -> anyhow::Result<R>
 where
-    F: FnOnce(&ResolvedConfig) -> anyhow::Result<R>,
+    F: FnOnce(&UpdateConfig) -> anyhow::Result<R>,
 {
-    if !config.is_initialized {
-        anyhow::bail!(UpdateError::ConfigNotInitialized);
+    match maybe_config {
+        Some(config) => f(&config),
+        None => anyhow::bail!(UpdateError::ConfigNotInitialized),
```
felangel, line 45: Can we add a test for this?

library/src/c_api.rs
eseidel: I think we do that already? Or am I misunderstanding.  As it notes in the diff, this is within "mod test"

library/src/config.rs
eseidel: Sure, on it.

eseidel:
Updated per comments. ptal.

bryanoltman (approved):

PR #45, opened 2023-06-26
fix: incorrect stack trace in shorebird code push

DK070202:
## Description

- Fixes incorrect stack trace in case of failure of `isNewPatchAvailableForDownload`, `currentPatchNumber`, and `nextPatchNumber`. 
- In case of failure of any of the above methods trace will only point to the `_runInIsolate` in place of the actual origin which is method itself.  
- Disabled the lint: `unnecessary_await_in_return` for this case. 
- Attaching trace generated by sample which mimicking exact conditions: 
- without await:  

  ```dart
    Future<void> isNewPatchAvailableForDownload() {
      return _runInIsolate();
    }

    Future<void> _runInIsolate() async {
      await Future.delayed(Duration(seconds: 1));
      throw Exception('Failure due to unknown reason');
    }
  ```
  
  ```txt
   Unhandled exception:
   Exception: Failure due to unknown reason
   #0      _runInIsolate
   trace.dart:11
   <asynchronous suspension>
  ```

- with await: 

  ```dart
   Future<void> isNewPatchAvailableForDownload() async {
     return await _runInIsolate();
   }

   Future<void> _runInIsolate() async {
     await Future.delayed(Duration(seconds: 1));
     throw Exception('Failure due to unknown reason');
   }
  ```
 
  ```txt
   Unhandled exception:
   Exception: Failure due to unknown reason
   #0      _runInIsolate
   trace.dart:11
   <asynchronous suspension>
   #1      isNewPatchAvailableForDownload
   trace.dart:6
   <asynchronous suspension>
  ```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore


eseidel:
Thank you very much for the patch!

DK070202:
> Yeah, unnecessary_await_in_return is such a foot-gun.
> 
> I would recommend we test this behavior if we want it to not break.

certainly! adding tests for it.

DK070202:
> Yeah, unnecessary_await_in_return is such a foot-gun.
> 
> I would recommend we test this behavior if we want it to not break.

Done!

felangel:
I think this can be closed because #55 refactored this and uncaught exceptions should bubble up along with their stacktraces.

Issue #49, opened 2023-07-05
ShorebirdCodePush should expose a "haveShorebird" function?


eseidel:
Some way for me to tell if it's connected to the Shorebird engine or not w/o reading the logs.

felangel:
Done in #55

PR #54, opened 2023-07-07
chore(shorebird_code_push): check for updater bindings when creating ShorebirdCodePush instance

bryanoltman:
## Description

Updates the shorebird_code_push package to replace the public `ShorebirdCodePush` constructor with a static `initialize` function that can return a null if no updater bindings are found.

Fixes https://github.com/shorebirdtech/updater/issues/49

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore


shorebird_code_push/lib/src/shorebird_code_push.dart
```diff
@@ -98,28 +92,19 @@ class ShorebirdCodePush {
...
     try {
       // Create a new Updater in the new isolate.
       return await Isolate.run(() => f(_buildUpdater()));
     } catch (error) {
-      _handleError(error);
-      return fallbackValue;
+      final logMessage = '$_loggingPrefix $error';
+      if (error is ArgumentError) {
+        // ffi function lookup failures manifest as ArgumentErrors.
+        throw ShorebirdCodePushNotAvailableException();
```
eseidel, line 104: I'm skeptical that we want code to run in `shorebird run` and then fail in `flutter run` by default.  It goes against the sell proposition that shorebird is "drop in".  I'm open to it, but I'm still not convinced.  Would love to learn more on monday.

eseidel (approved):

shorebird_code_push/lib/src/shorebird_code_push.dart
bryanoltman: I'm not 100% sold on it either - I'll think about it some more over the weekend. This is largely a response to people wanting more info from the errors (see the PR this is based on) and a helping a lot of people with "I made a patch, why isn't it downloading" issues this week.

eseidel:
I think your idea of that maybe updater is nullable and init fails when
shorebird isn’t there is a good one. Would make people guard against
absence without needing everything to throw

shorebird_code_push/lib/src/shorebird_code_push.dart
bryanoltman: Thought about this some more and came up with (what I think is) a better idea. I've made the `ShorebirdCodePush` constructor private and introduced a static `initialize` function that returns a non-null `ShorebirdCodePush` if our updater bindings are present or `null` if they are not.

shorebird_code_push/example/lib/main.dart
```diff
@@ -40,34 +42,51 @@ class MyHomePage extends StatefulWidget {
...
   void initState() {
     super.initState();
-    // Request the current patch number.
-    _shorebirdCodePush.currentPatchNumber().then((currentPatchVersion) {
-      if (!mounted) return;
-      setState(() {
-        _currentPatchVersion = currentPatchVersion;
+    if (_isShorebirdAvailable) {
+      // Request the current patch number.
+      _shorebirdCodePush!.currentPatchNumber().then((currentPatchVersion) {
```
felangel, line 52: I don't love this personally because you have to always deal with a nullable instance and sprinkle null checks throughout. What are you thoughts on having two implementations of the `ShorebirdCodePush` interface and when you don't have shorebird integrated you get no-ops/errors.

bryanoltman: > I'm not 100% clear on what the problem we're solving is tbh.

There have been several Discord questions over the past week where users expressed confusion about the behavior of this plugin due to the silent failures.

bryanoltman: Is that much different than checking whether `shorebirdCodePush` is null? I feel like it just shifts the null check to the return value instead of the `shorebirdCodePush` instance.

felangel:
I think this can be closed in favor of #55?

felangel:
Closing this for now but happy to address anything that might've been missed in #55 👍

PR #55, opened 2023-07-10
refactor(shorebird_code_push): ffi vs noop implementations

### LINKED ISSUE #57: Not able to build web due to dart import ffi.
Im developing a web and mobile app and integrated shorebird code push in mobile app and its working very well. only issue is I'm not able to build web as it throws error of import 'dart:ffi'.

Need solution for this. @eseidel

felangel:
## Description

- refactor(shorebird_code_push): ffi vs noop implementations
- fix(shorebird_code_push): should run on web (closes #57)
- feat(shorebird_code_push): expose `isShorebirdAvailable` API

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [X] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
+++ b/shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
@@ -8,7 +8,8 @@ import 'package:shorebird_code_push/src/updater.dart';
 /// {@endtemplate}
 class ShorebirdCodePushFfi implements ShorebirdCodePushBase {
   /// {@macro shorebird_code_push}
-  ShorebirdCodePushFfi({Updater? updater}) : _updater = updater ?? Updater();
+  ShorebirdCodePushFfi({Updater? updater})
+      : _updater = updater ?? const Updater();
 
   final Updater _updater;
 
--- a/shorebird_code_push/test/src/updater_test.dart
+++ b/shorebird_code_push/test/src/updater_test.dart
@@ -13,7 +13,7 @@ void main() {
     setUp(() {
       updaterBindings = _MockUpdaterBindings();
 
-      updater = Updater();
+      updater = const Updater();
       Updater.bindings = updaterBindings;
     });
 


felangel:
> I worry this might be more confusing for users than the current state of affairs, as the no-op version doesn't provide any feedback that shorebird is unavailable. I understand the desire to have this plugin just work with `flutter run`, but I think we need some way to communicate to users (without relying on them having fully read the docs) why the plugin is providing "wrong" results.

Imo it should be obvious if APIs like `currentPatchNumber` return null when shorebird is not available and non-null values when shorebird is available 🤷‍♂️

shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @override
+  Future<bool> isNewPatchReadyToInstall() async {
+    final patchNumbers =
+        await Future.wait([currentPatchNumber(), nextPatchNumber()]);
+    final currentPatch = patchNumbers[0];
+    final nextPatch = patchNumbers[1];
+
+    return nextPatch != null && currentPatch != nextPatch;
```
bryanoltman, line 53: Next patch and current patch will be equal if there's no new update iirc

shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @override
+  bool isShorebirdAvailable() => true;
+
+  /// Creates an [Updater] in a separate isolate and runs the given function.
+  Future<T> _runInIsolate<T>(T Function(Updater updater) f) async {
+    return Isolate.run(() => f(_updater));
```
bryanoltman, line 61: Do we not need to reconstruct the updater in the isolate? Iirc, there was some sort of bindings failure (sorry for the lack of specifics) when using an updater created in a different isolate

shorebird_code_push/lib/shorebird_code_push_io.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  @override
+  Future<int?> currentPatchNumber() => _delegate.currentPatchNumber();
+
+  @override
+  Future<int?> nextPatchNumber() => _delegate.nextPatchNumber();
+
+  @override
+  Future<void> downloadUpdateIfAvailable() {
+    return _delegate.downloadUpdateIfAvailable();
```
bryanoltman, line 39: Fat arrow syntax here?

shorebird_code_push/lib/shorebird_code_push.dart
```diff
@@ -1,4 +1,5 @@
 /// Get info about your Shorebird code push app
 library shorebird_code_push;
 
-export 'src/shorebird_code_push.dart';
+export 'shorebird_code_push_io.dart'
```
felangel, line 4: fixes #57

shorebird_code_push/lib/shorebird_code_push_io.dart
felangel: I did this intentionally because fat-arrow wraps for this line.

shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
felangel: This should still be correct though right? `nextPatch` will only be non-null when an update has been installed. I didn't change this logic btw.

shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
felangel: I don't think it's necessary -- I tested locally and it works as expected.

shorebird_code_push/lib/shorebird_code_push_io.dart
```diff
@@ -0,0 +1,46 @@
...
+/// {@endtemplate}
+class ShorebirdCodePush implements ShorebirdCodePushBase {
+  /// {@macro shorebird_code_push}
+  ShorebirdCodePush() {
+    try {
+      // If the Shorebird Engine is not available, this will throw an exception.
+      final updater = Updater()..currentPatchNumber();
+      _delegate = ShorebirdCodePushFfi(updater: updater);
+    } catch (error) {
+      _delegate = ShorebirdCodePushNoop();
```
eseidel, line 17: Do we need to log?

shorebird_code_push/lib/shorebird_code_push_io.dart
eseidel: Maybe it's time to start a style guide. 🙃  (I'm mostly kidding.)

shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
```diff
@@ -0,0 +1,57 @@
...
+  @override
+  Future<bool> isNewPatchAvailableForDownload() {
+    return Isolate.run(_updater.checkForUpdate);
+  }
+
+  @override
+  Future<int?> currentPatchNumber() {
+    return Isolate.run(() {
+      final currentPatchNumber = _updater.currentPatchNumber();
+      return currentPatchNumber == 0 ? null : currentPatchNumber;
```
eseidel, line 24: I'd use a constant or at least a comment to explain the magic value 0.

shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
```diff
@@ -0,0 +1,57 @@
...
+    );
+  }
+
+  @override
+  Future<void> downloadUpdateIfAvailable() async {
+    await Isolate.run(_updater.downloadUpdate);
+  }
+
+  @override
+  Future<bool> isNewPatchReadyToInstall() async {
```
eseidel, line 44: This is fine, but we also could just add this direct method on the library if it's needed often.

eseidel: The behavior here is also more of telling the caller that the next boot will be a different version.  The patch is already installed at this point.

eseidel (approved):
Bryan should see, since he was working in this area, but we can also add any of his changes after landing too.  lgtm!

shorebird_code_push/lib/src/shorebird_code_push_ffi.dart
felangel: This API was already part of the bindings, I just moved it into a different file/reorganized things a bit.

shorebird_code_push/lib/shorebird_code_push_io.dart
felangel: updated ✅

Issue #56, opened 2023-07-10
Better error messaging for hash mismatch errors


bryanoltman:

felangel:
Is this needed anymore? We've added the `shorebird preview` command and deprecated `shorebird run` so it be a lot less likely for folks to encounter this error now.

bryanoltman:
We can maybe punt on it and see how many reports of hash mismatches we get over the week or two after `preview` is released?

bryanoltman:
But yeah, I think if `preview` makes this go away entirely, we don't need to worry about sprucing this up.

eseidel:
I suspect this will happen as part of us cleaning up the rust logging.  Right now rust logs a wall of text when it doesn't need to anymore.

eseidel:
Got another report of this from a customer tonight.

PR #63, opened 2023-07-31
fix: Improve rust logging and error clarity

eseidel:
https://github.com/shorebirdtech/shorebird/issues/779, https://github.com/shorebirdtech/updater/issues/61

* moved most info! logs to be debug! logs: https://github.com/shorebirdtech/shorebird/issues/779
* Changed the default logging level to info rather than debug on Android.
* Added an explicit check for "file not found" error to avoid logging in that (expected) case for state.json creation.
* Fixed check_hash to return Ok() or Err(), the bool was unnecessary.  Also added better logging: https://github.com/shorebirdtech/updater/issues/56
* Made updater thread print errors: https://github.com/shorebirdtech/updater/issues/61
* Deleted unused assets.rs


eseidel:
The coverage "going down" is because our coverage bot currently gets confused dart vs. rust.  We need to fix to upload both.  We have 100% coverage on Dart and 92% Rust and whichever ran last wins.

felangel (approved):

PR #64, opened 2023-08-02
test: improve rust coverage

eseidel:
Refactored our android code slightly so we could test all the
branches of our "find the right split apk" function.

This may have been a waste of time.

### DIFF
--- a/library/src/cache.rs
+++ b/library/src/cache.rs
@@ -478,6 +478,32 @@ mod tests {
         // Calling a second time should not error.
         state.mark_patch_as_bad(number);
     }
+
+    #[test]
+    fn do_not_mark_bad_patch_good() {
+        let tmp_dir = TempDir::new("example").unwrap();
+        let mut state = test_state(&tmp_dir);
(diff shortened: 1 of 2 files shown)


felangel (approved):

library/src/cache.rs
```diff
@@ -478,6 +478,32 @@ mod tests {
...
         state.mark_patch_as_bad(number);
     }
+
+    #[test]
+    fn do_not_mark_bad_patch_good() {
+        let tmp_dir = TempDir::new("example").unwrap();
+        let mut state = test_state(&tmp_dir);
+        let bad_patch = fake_patch(&tmp_dir, 1);
+        state.mark_patch_as_bad(bad_patch.number);
+        state.mark_patch_as_good(bad_patch.number);
```
bryanoltman, line 488: Should this fail in some non-silent way?

eseidel: Certainly could (at least at the `updater` layer), that's not a bad idea.  Then the shorebird_* c layer could just turn that into a message.

PR #65, opened 2023-08-02
feat: add support for parsing update_behavior from shorebird.yaml

eseidel:
The actual support will be in the C++ engine, but this keeps
all yaml parsing in the Rust code for simplicity.

### DIFF
--- a/library/include/updater.h
+++ b/library/include/updater.h
@@ -14,6 +14,12 @@
 #endif
 
 
+typedef enum UpdateBehavior {
+  BackgroundOnLaunch,
+  Manual,
+  WaitOnLaunch,
+} UpdateBehavior;
+
 /**
  * Struct containing configuration parameters for the updater.
  * Passed to all updater functions.
@@ -54,6 +60,8 @@ extern "C" {
 SHOREBIRD_EXPORT
 bool shorebird_init(const struct AppParameters *c_params,
                     const char *c_yaml);
+
+SHOREBIRD_EXPORT enum UpdateBehavior shorebird_update_behavior(void);
 
 /**
  * The currently running patch number, or 0 if the release has not been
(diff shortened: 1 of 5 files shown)


eseidel:
https://github.com/shorebirdtech/engine/pull/40 is the C++ side of this.

library/include/updater.h
```diff
@@ -14,6 +14,12 @@
 #endif
 
 
+typedef enum UpdateBehavior {
+  BackgroundOnLaunch,
+  Manual,
+  WaitOnLaunch,
```
felangel, line 20:
```suggestion
  AsyncOnLaunch,
  SyncOnLaunch,
  Manual,
```

felangel: feel free to ignore but tried to come up with more consistent names for the modes 🤷

eseidel: Sure. 🤷   Naming is the hardest part here.  Those are the internal enum names, but we also have to figure out the public yaml names (and the yaml property name).

felangel (approved):

library/include/updater.h
felangel: Yeah the public facing names are more important especially because it'll be tough to change after 🤷

felangel: So just to make sure I follow the modes we want to support are:

1. Check for updates on launch and install in background (`background`, what we currently do)
2. Check for updates on launch and install blocking (`wait`)
3. Do not check for updates, leave it up to the developer (`manual`)

felangel: What if we split this into `update_check_behavior` and `update_installation_behavior`?

```yaml
update_check_behavior: launch
update_install_behavior: blocking | non-blocking
```

felangel: Also the more I think about it the more I wonder whether or not we should just have a flag:

```yaml
auto_update: true | false
```

For all custom scenarios we could just leave it up to `package:shorebird_code_push`

PR #70, opened 2023-08-10
feat: add client_id to UpdaterState

bryanoltman:
## Description

Adds a client_id field to UpdaterState, the contents of which will be a random UUID that is generated once, either on load (for existing clients) or on UpdaterState creation (for new clients).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache.rs
+++ b/library/src/cache.rs
@@ -137,7 +137,7 @@ impl UpdaterState {
         Ok(state)
     }
 
-    fn create_and_save_new(cache_dir: &Path, release_version: &str) -> Self {
+    fn create_new_and_save(cache_dir: &Path, release_version: &str) -> Self {
         let state = Self::new(cache_dir.to_owned(), release_version.to_owned());
         let _ = state.save();
         state
@@ -152,12 +152,12 @@ impl UpdaterState {
                         "release_version changed {} -> {}, clearing updater state",
                         loaded.release_version, release_version
                     );
-                    return Self::create_and_save_new(cache_dir, release_version);
+                    return Self::create_new_and_save(cache_dir, release_version);
                 }
                 let validate_result = loaded.validate();
                 if let Err(e) = validate_result {
                     warn!("Error while validating state: {:#}, clearing state.", e);
-                    return Self::create_and_save_new(cache_dir, release_version);
+                    return Self::create_new_and_save(cache_dir, release_version);
                 }
                 loaded
             }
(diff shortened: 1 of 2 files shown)


library/src/cache.rs
```diff
@@ -121,7 +128,11 @@ impl UpdaterState {
         let reader = BufReader::new(file);
         // TODO: Now that we depend on serde_yaml for shorebird.yaml
         // we could use yaml here instead of json.
-        let state = serde_json::from_reader(reader)?;
+        let mut state: UpdaterState = serde_json::from_reader(reader)?;
+        if state.client_id.is_none() {
+            // Generate a client id if we don't already have one.
+            state.client_id = Some(generate_client_id());
```
felangel, line 134: when does the next save happen? Are we guaranteed to persist the newly generated client_id? Is it possible for concurrent checks to generate separate client_ids?

bryanoltman: > when does the next save happen?
It depends on which function loads this. I added a `save` here to make this easier to reason about.

> Are we guaranteed to persist the newly generated client_id?
Added a `create_and_save_new` function to UpdaterState that is called in `load_or_new_on_error` when we construct a new UpdaterState.

> Is it possible for concurrent checks to generate separate client_ids?
No – the `with_config` wrapper acquires a lock before reading/writing state.

library/src/cache.rs
```diff
@@ -121,10 +128,21 @@ impl UpdaterState {
...
+            let _ = state.save();
+        }
         Ok(state)
     }
 
+    fn create_and_save_new(cache_dir: &Path, release_version: &str) -> Self {
```
felangel, line 140:
```suggestion
    fn new_and_save(cache_dir: &Path, release_version: &str) -> Self {
```
nit

bryanoltman: I tried to follow the naming pattern that exists in the `UpdaterState`. There's a lot of verb-noun naming for functions that do more than retrieve values (`validate_slot`, `activate_latest_bootable_patch``set_next_boot_patch_slot`, etc.), so I updated this to `create_new_and_save` to keep the `create` verb in there.

eseidel (commented):
I guess there is some risk that callers call new() instead of new_and_save() and end up generating uuid's twice, but seems small.

lgtm.  we may need to update our privacy policy and guidance for store submissions.

PR #71, opened 2023-08-10
feat: add support for reporting patch install events

bryanoltman:
## Description

Adds the PatchInstallEvent type and posts to the `api/v1/patches/events` endpoint when the engine reports a successful launch and we mark a patch as good.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/network.rs
+++ b/library/src/network.rs
@@ -18,8 +18,13 @@ fn patches_check_url(base_url: &str) -> String {
     return format!("{}/api/v1/patches/check", base_url);
 }
 
+fn patches_events_url(base_url: &str) -> String {
+    return format!("{}/api/v1/patches/events", base_url);
+}
+
 pub type PatchCheckRequestFn = fn(&str, PatchCheckRequest) -> anyhow::Result<PatchCheckResponse>;
 pub type DownloadFileFn = fn(&str) -> anyhow::Result<Vec<u8>>;
+pub type PatchInstallSuccessFn = fn(&str, PatchInstallEvent) -> anyhow::Result<()>;
 
 /// A container for network clalbacks which can be mocked out for testing.
 #[derive(Clone)]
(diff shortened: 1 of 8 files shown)


library/src/network.rs
```diff
@@ -162,15 +248,39 @@ pub fn send_patch_check_request(
         platform: current_platform().to_string(),
         arch: current_arch().to_string(),
     };
-    debug!("Sending patch check request: {:?}", request);
```
felangel, line 165: why did we remove the debug logs?

library/src/updater.rs
```diff
@@ -344,7 +345,7 @@ pub fn report_launch_start() -> anyhow::Result<()> {
         let mut state =
             UpdaterState::load_or_new_on_error(&config.cache_dir, &config.release_version);
         // Validate that we have an installed patch.
-        // Make that patch the "booted" patch.
+        // Make that patch is the "booted" patch.
```
felangel, line 348:
```suggestion
        // Make that patch the "booted" patch.
```

I think it was correct before?

library/src/network.rs
bryanoltman: My mistake, thought I added those in my testing :)

library/src/network.rs
```diff
@@ -171,6 +257,32 @@ pub fn send_patch_check_request(
...
+        client_id.to_string(),
+        patch_number,
+        current_platform().to_string(),
+        config.release_version.clone(),
+    );
+    let request = CreatePatchInstallEventRequest { event };
+
+    let patch_install_success_fn = config.network_hooks.patch_install_success_fn;
+    let url = &patches_events_url(&config.base_url);
+    patch_install_success_fn(url, request)
```
felangel, line 283:
```suggestion
    let url = &patches_events_url(&config.base_url);
    config.network_hooks.patch_install_success_fn(url, request)
```
nit but would this work?

felangel (approved):

library/src/network.rs
bryanoltman: It would need to be `(config.network_hooks.patch_install_success_fn)(url, request)`, but yeah, that would work. I did it this way for consistency with the other functions in that file that use network hooks.

PR #73, opened 2023-08-31
fix: show patch check info in logs

eseidel:
We directed users to look for this log, but it was no longer visible after my recent changes to reduce logging, so adding back.
https://docs.shorebird.dev/troubleshooting#how-to-fix-it-1

I don't know how to test in rust for log calls, so just leaving
a comment for now.

### DIFF
--- a/library/src/network.rs
+++ b/library/src/network.rs
@@ -13,7 +13,7 @@ use crate::config::{current_arch, current_platform, UpdateConfig};
 
 // https://stackoverflow.com/questions/67087597/is-it-possible-to-use-rusts-log-info-for-tests
 #[cfg(test)]
-use std::println as debug; // Workaround to use println! for logs.
+use std::{println as info, println as debug}; // Workaround to use println! for logs.
 
 fn patches_check_url(base_url: &str) -> String {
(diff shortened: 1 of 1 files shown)


felangel (approved):

library/src/network.rs
```diff
@@ -13,7 +13,7 @@ use crate::config::{current_arch, current_platform, UpdateConfig};
 
 // https://stackoverflow.com/questions/67087597/is-it-possible-to-use-rusts-log-info-for-tests
 #[cfg(test)]
-use std::println as debug; // Workaround to use println! for logs.
+use std::{println as info, println as debug}; // Workaround to use println! for logs.
```
bryanoltman, line 16: Is this line necessary since we're changing the "sending patch check request" log to an `info!`?

bryanoltman (approved):

library/src/network.rs
eseidel: Correct.  Previously info! wasn't used in this file.

bryanoltman: Ah! LGTM then.

PR #75, opened 2023-09-06
refactor: move shorebird_code_push_io and _web into src/

eseidel:
Users were sometimes importing the _web version thinking it
was the "web safe" import, rather than shorebird_code_push.dart
and then being surprised when ShorebirdCodePush did nothing.

Fixes https://github.com/shorebirdtech/shorebird/issues/1209


eseidel:
Clearly I need to wire up coverage flags for this repo too.

PR #76, opened 2023-09-07
fix: don't regenerate client id if UpdaterState successfully loaded

bryanoltman:
## Description

Fixes issue where client_id was being regenerated when the app's release version changed or when the previous state failed to validate.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache.rs
+++ b/library/src/cache.rs
@@ -59,11 +59,11 @@ pub struct UpdaterState {
 }
 
 impl UpdaterState {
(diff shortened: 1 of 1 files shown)


eseidel (approved):
Patch is fine, I'm not sure it's worth the hassle (since we don't currently use the client_id do we?) and it's just more functionality to maintain.  But this does make the client_id more stable (as someone might expect) and does add more testing so 🤷

PR #78, opened 2023-09-07
feat: report patch install failure events

eseidel:
This doesn't tell us anything about the failure, just that it
happened.  Which at least will let us communicate that to customers.

This currently has no tests and will need some before landing.

### DIFF
--- a/library/src/events.rs
+++ b/library/src/events.rs
@@ -3,7 +3,7 @@
 
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum EventType {
     PatchInstallSuccess,
     PatchInstallFailure,
--- a/library/src/events.rs
+++ b/library/src/events.rs
@@ -1,12 +1,12 @@
 // This file's job is to deal with the update_server and network side
 // of the updater library.
 
-use serde::{Serialize, Serializer};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum EventType {
     PatchInstallSuccess,
-    // PatchInstallFailure,
+    PatchInstallFailure,
 }
 
 impl Serialize for EventType {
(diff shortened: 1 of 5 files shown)


library/src/events.rs
```diff
@@ -16,16 +16,32 @@ impl Serialize for EventType {
     {
         serializer.serialize_str(match self {
             EventType::PatchInstallSuccess => "__patch_install__",
-            // EventType::PatchInstallFailure => "__patch_install_failure__",
+            EventType::PatchInstallFailure => "__patch_install_failure__",
         })
     }
 }
 
+impl<'de> Deserialize<'de> for EventType {
```
bryanoltman, line 24: Mostly for my learning: I know it's valid to give lifetime annotationss whatever name you'd like, but I most commonly see `'a`. Are there best practices around when/what to name them different things?

library/src/updater.rs
```diff
@@ -223,11 +221,37 @@ fn update_internal(_: &UpdaterLockState) -> anyhow::Result<UpdateStatus> {
     // Saves state to disk (holds Config lock while writing).
 
     let config = copy_update_config()?;
+    // We should never try to write this state as some other writer may be
+    // racing with us, we should get a new state inside a lock if we want
+    // to write.
+    let read_only_state =
+        UpdaterState::load_or_new_on_error(&config.cache_dir, &config.release_version);
+
+    // We discard any events if we have more than 3 queued to make sure
```
bryanoltman, line 230: What's the significance of 3? Why not 2 or 4?

library/src/updater.rs
```diff
@@ -223,11 +221,37 @@ fn update_internal(_: &UpdaterLockState) -> anyhow::Result<UpdateStatus> {
...
+            error!("Failed to report event: {:?}", err);
+        }
+    }
+    // We're abusing the config lock as a UpdateState lock for now.
+    let read_only_state = with_config(|_| {
+        let mut state =
+            UpdaterState::load_or_new_on_error(&config.cache_dir, &config.release_version);
+        // This will clear any events which got queued between the time we
+        // loaded the state now, but that's OK for now.
+        let result = state.clear_events();
```
bryanoltman, line 245: How does this change get saved?

library/src/updater.rs
```diff
@@ -367,7 +393,22 @@ pub fn report_launch_failure() -> anyhow::Result<()> {
                 )))?;
         // Ignore the error here, we'll try to activate the next best patch
         // even if we fail to mark this one as bad (because it was already bad).
-        let _ = state.mark_patch_as_bad(patch.number);
+        let mark_result = state.mark_patch_as_bad(patch.number);
```
bryanoltman, line 396: Same Q here re: saving state

library/src/events.rs
eseidel: It's just matching what the docs do for this specific class. 🤷  https://serde.rs/impl-deserialize.html

library/src/updater.rs
eseidel: No significance.  We could do 2 or 4 or 1 or whatever.  Originally I used 10 🤷

library/src/updater.rs
eseidel:
```    pub fn clear_events(&mut self) -> Result<()> {
        self.queued_events.clear();
        self.save()
    }```

library/src/updater.rs
eseidel: It gets accidentally saved below by the queue_patch call.

eseidel: Sorry by the activate_latest_bootable_patch() call.

library/src/updater.rs
bryanoltman: Oh interesting. I wonder if the "save at the end of a mutating method" pattern is something we want to either adopt throughout or discourage, as it's currently done inconsistently in `UpdaterState`.

eseidel: Yeah, the inconsistency is the problem.  I suspect what we actually want is some sort of file system object which you have to get a lock to use?  And then UpdaterState doesnt' know how to save itself, rather it gets passed to something that knows how to read or write it?

bryanoltman (approved):

library/src/updater.rs
eseidel: I added a doc about where the save happens.  Would you like others?  Happy to, just would like inspiration as to what docs you're seeking.

eseidel:
I think we/I should take a separate pass to make the saving more explicit/sane.

PR #80, opened 2023-09-08
chore: run `cargo clippy` over our rust codebase

eseidel:
I ran `cargo clippy`.

I've not tried to configure it or make it run on PRs yet, but this
is a start.

### DIFF
--- a/library/src/android.rs
+++ b/library/src/android.rs
@@ -73,7 +73,7 @@ fn android_arch_names() -> &'static ArchNames {
         apk_split: "armeabi_v7a", // e.g. base-armeabi_v7a.apk
         lib_dir: "armeabi-v7a",   // e.g. lib/armeabi-v7a/libapp.so
     };
-    return &ARCH;
+    &ARCH
 }
 
 // This is public so c_api can use this for testing.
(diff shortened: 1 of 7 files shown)


library/src/android.rs
```diff
@@ -184,9 +184,7 @@ pub(crate) fn open_base_lib(apks_dir: &Path, lib_name: &str) -> anyhow::Result<C
     Ok(Cursor::new(buffer))
 }
 
-pub fn libapp_path_from_settings(
-    original_libapp_paths: &Vec<String>,
-) -> Result<PathBuf, UpdateError> {
+pub fn libapp_path_from_settings(original_libapp_paths: &[String]) -> Result<PathBuf, UpdateError> {
```
eseidel, line 187: Turns out it's better to pass a slice, rather than cause it to copy the Vec.

library/src/c_api.rs
```diff
@@ -173,8 +173,12 @@ pub extern "C" fn shorebird_next_boot_patch_path() -> *mut c_char {
 }
 
 /// Free a string returned by the updater library.
+/// # Safety
+///
+/// If this function is called with a non-null pointer, it must be a pointer
+/// returned by the updater library.
 #[no_mangle]
-pub extern "C" fn shorebird_free_string(c_string: *mut c_char) {
+pub unsafe extern "C" fn shorebird_free_string(c_string: *mut c_char) {
```
eseidel, line 181: I think most of our c_api is unsafe... just isn't caught by clippy, yet.

bryanoltman (approved):

PR #81, opened 2023-09-08
chore: fix 30 of the 54 warnings from clippy pedantic.

eseidel:
I just ran `cargo clippy -- -W clippy::pedantic` and fixed things.

These are more invasive that the default set and the remaining
warnings are mostly about our (abysmal) public docs missing
Error and Panic sections to explain errors and panicks.


eseidel:
I'm not actually sure if this is worth doing.  I did learn a bit about rust and markdown though. 🤷

bryanoltman (approved):

PR #87, opened 2023-09-20
fix: don't serialize UpdaterState cache_dir as it can change between app runs

bryanoltman:
## Description

Patches are sometimes failing to apply on iOS due to the reasons explained [here](https://github.com/shorebirdtech/shorebird/issues/1296#issuecomment-1728308059). In short, UpdaterState is saving its location, which is not always guaranteed to be accessible to the app, as iOS can move the app's home directory between runs.

Fixes https://github.com/shorebirdtech/shorebird/issues/1296

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache.rs
+++ b/library/src/cache.rs
@@ -41,7 +41,9 @@ struct Slot {
 #[derive(Debug, Deserialize, Serialize)]
 pub struct UpdaterState {
     // Per-device state:
-    /// Where this writes to disk.
+    /// Where this writes to disk. Don't serialize this field, as it can change
+    /// between runs of the app.
+    #[serde(skip)]
     cache_dir: PathBuf,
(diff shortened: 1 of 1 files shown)


eseidel (approved):
I think this test is great, but I think you should also call one of the methods which gets a path to a slot and verify it's a new path.

Similarly the download dir path?

bryanoltman:
@eseidel Updated to test the patch path. `download_dir` is not written to disk – it's derived from the `AppParameters` provided by the engine.

felangel (approved):
LGTM but @eseidel is a better reviewer

eseidel (approved):

PR #88, opened 2023-09-21
feat: treat iOS patches as diffs instead of full artifacts

bryanoltman:
## Description

Updates `prepare_for_install` function to treat iOS patch artifacts as binary diffs instead of complete artifacts.

Also introduces the `platform` module to more cleanly separate iOS- and Android-specific code.

Note that this will require changes in the CLI, which are still WIP.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


library/src/c_api.rs
```diff
@@ -254,7 +254,10 @@ pub extern "C" fn shorebird_report_launch_success() {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::network::{testing_set_network_hooks, PatchCheckResponse};
+    use crate::{
+        network::{testing_set_network_hooks, PatchCheckResponse},
```
eseidel, line 258: I don't know what preferred rust style is, but I've tended to just split these out into two separate lines.
use crate::network::{testing_set_network_hooks, PatchCheckResponse};
use crate::platform;

library/src/platform/unknown.rs
```diff
@@ -0,0 +1,20 @@
...
+use anyhow::bail;
+use std::{
+    io::Cursor,
+    path::{Path, PathBuf},
+};
+
+const UNKNOWN_PLATFORM_ERR_MSG: &str = "Unknown platform";
+
+pub fn open_base_lib(_app_dir: &Path, _lib_name: &str) -> anyhow::Result<Cursor<Vec<u8>>> {
+    bail!(UNKNOWN_PLATFORM_ERR_MSG)
```
eseidel, line 11: Do the tests still run on a linux host, etc with this?  They must...

eseidel (approved):
Seems a bit cutesy, but sure. 🤷

library/src/c_api.rs
bryanoltman: 🤷  I just did what the formatter/vscode put in there. Will update.

library/src/platform/unknown.rs
bryanoltman: Yep, they do. The same tests still run (i.e., the Android tests). iOS tests don't currently execute, which, in addition to coverage, is why this is a draft.

This particular code never runs, it's solely so that the package will build in a non-test, non-android, non-ios environment.

library/src/c_api.rs
bryanoltman: It looks like it's still in an unstable state, but `rustfmt` has an option to [do this for us](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=group#imports_granularity). Once that's stable, we should add it to `rustfmt.toml`.

eseidel:
Yeah, I've not really explored the world of fmt options or linting at all in Rust.  I'm sure there is a ton of useful stuff we could turn on.  I should dig up Fuchsia's rust config files.

PR #90, opened 2023-09-26
chore: Revert "feat: treat iOS patches as diffs instead of full artifacts (#88)"

bryanoltman:
This reverts commit f4db98b7597d4b6d2ca162d8b97d98cb78e964fe.

Because this approach will not work for iOS artifacts that are signed after release artifact is submitted to shorebird (as happens in our iOS add-to-app setup, where the .xcframework is signed on embed).


eseidel (approved):
I look forward to this patch's triumphant return.

PR #92, opened 2023-09-27
feat: add script to relink .a as a .o with exported symbols list

eseidel:
This should help us hide all the extra rust symbols when importing libupdater.a into libflutter.so.


eseidel:
I haven't successfully seen it link yet (hence why a draft).

eseidel:
OK, this works now and saves 2mb on iOS (still 6mb more to go).

Testing on Android.

eseidel:
Closing until we start measuring: https://github.com/shorebirdtech/docs/issues/19

PR #93, opened 2023-10-02
refactor: move patch management functionality out of UpdaterState 

bryanoltman:
## Description

Restructures UpdaterState by separating its responsibilities into separate `impl`s and by moving patch management into a separate PatchManagement module.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


library/src/updater.rs
```diff
@@ -370,33 +370,34 @@ pub fn current_boot_patch() -> anyhow::Result<Option<PatchInfo>> {
...
 /// This will mark the patch as bad and activate the next best patch.
 pub fn report_launch_failure() -> anyhow::Result<()> {
     info!("Reporting failed launch.");
+
     with_config(|config| {
         let mut state =
             UpdaterState::load_or_new_on_error(&config.storage_dir, &config.release_version);
 
+        // Attempting to get next_boot_patch might return None if the failure was due to
+        // the patch being altered on disk.
```
bryanoltman, line 391: Made a decision not to report patches that fail due to local disk changes. This felt reasonable to me, as there is nothing the app developer can do in this case.

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,674 @@
...
+    /// safe to use for future boots.
+    fn record_boot_success_for_patch(&mut self, patch_number: usize) -> Result<()>;
+
+    /// Records that the patch with number patch_number failed to boot, and ensures
+    /// that it will never be returned as the next boot or last booted patch.
+    fn record_boot_failure_for_patch(&mut self, patch_number: usize) -> Result<()>;
+
+    /// The highest patch number that has been added. This may be higher than the
+    /// last booted or next boot patch if we downloaded a patch that failed to boot.
+    fn highest_seen_patch_number(&self) -> Option<usize>;
```
bryanoltman, line 62: This functions differently than the previous implementation.

Before: represented the latest patch we still have on disk so as to prevent re-downloading patches we already have
Now: the last patch we've seen to prevent redownloading patches we've already downloaded. This means that a bad patch will not be redownloaded/reinstalled, which may or may not be what we want (this might cause problems in cases where a patch becomes bad due to corruption on disk)

library/src/updater.rs
eseidel: I guess it comes down to how it manifests (does the user notice?).  A patch failure is still a patch failure and probably worth reporting, possibly with context of some form?

library/src/cache/disk_io.rs
```diff
@@ -0,0 +1,104 @@
...
+    let path_as_ref = path.as_ref();
+    let containing_dir = path_as_ref
+        .parent()
+        .with_context(|| format!("Failed to get parent dir for {:?}", path_as_ref))?;
+
+    // Because File::create can sometimes fail if the full directory path doesn't exist,
+    // we create the directories in its path first.
+    std::fs::create_dir_all(containing_dir).with_context(|| {
+        format!(
+            "Failed to create root dir {:?} when initializing PatchManager",
```
eseidel, line 29: This isn't in PatchManager.

library/src/cache/disk_io.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    let path_as_ref = path.as_ref();
+    let containing_dir = path_as_ref
+        .parent()
+        .with_context(|| format!("Failed to get parent dir for {:?}", path_as_ref))?;
+
+    // Because File::create can sometimes fail if the full directory path doesn't exist,
+    // we create the directories in its path first.
+    std::fs::create_dir_all(containing_dir).with_context(|| {
+        format!(
```
eseidel, line 28: CodeCov says you missed the failure case.  Might be worth testing?

eseidel: Probably can just create a directory or path in the way and drop write permissions on it?

library/src/cache/disk_io.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    let path_as_ref = path.as_ref();
+    if !path_as_ref.exists() {
+        bail!("File {} does not exist", path_as_ref.display());
+    }
+
+    let file = File::open(path_as_ref)?;
+    let reader = BufReader::new(file);
+    serde_json::from_reader(reader)
+        .context(format!("failed to deserialize from {:?}", &path_as_ref))
```
eseidel, line 55: with_context will delay creation of the string if you think it matters.  context() will do it eagerly.

library/src/cache/disk_io.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    #[derive(Serialize, Deserialize, PartialEq, Eq)]
+    struct TestStruct {
+        a: u32,
+        b: String,
+    }
+
+    #[test]
+    fn writes_and_reads_serialized_object() -> Result<()> {
+        let test_struct = TestStruct {
```
eseidel, line 75: I love that we have a TestStruct!

library/src/cache/disk_io.rs
```diff
@@ -0,0 +1,104 @@
...
+    use anyhow::{Ok, Result};
+
+    #[derive(Serialize, Deserialize, PartialEq, Eq)]
+    struct TestStruct {
+        a: u32,
+        b: String,
+    }
+
+    #[test]
+    fn writes_and_reads_serialized_object() -> Result<()> {
```
eseidel, line 74: Why does a Test return a Result?  Is that a thing?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,683 @@
...
+
+// https://stackoverflow.com/questions/67087597/is-it-possible-to-use-rusts-log-info-for-tests
+#[cfg(test)]
+use std::{println as info, println as error}; // Workaround to use println! for logs.
+
+const PATCHES_DIR_NAME: &str = "patches";
+const PATCHES_STATE_FILE_NAME: &str = "patches_state.json";
+
+#[derive(Copy, Clone, Debug, Deserialize, Serialize, PartialEq, Eq, Hash)]
+struct PatchMetadata {
```
eseidel, line 20: So is the patch just one file or is it a directory?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// Copies the patch file at file_path to the manager's directory structure sets
+    /// this patch as the next patch to boot.
+    fn add_patch(&mut self, number: usize, file_path: &Path) -> Result<()>;
+
+    /// Returns the patch we most recently successfully booted from (usually the currently running patch),
+    /// or None if no patch is installed.
+    fn last_successfully_booted_patch(&self) -> Option<PatchInfo>;
+
+    /// Returns the next patch to boot, or None if no new patch has been downloaded.
+    fn get_next_boot_patch(&mut self) -> Option<PatchInfo>;
```
eseidel, line 55: I don't know rust style.  Does rust use get_ in names?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,683 @@
...
+
+    /// Records that the patch with number patch_number booted successfully and is
+    /// safe to use for future boots.
+    fn record_boot_success_for_patch(&mut self, patch_number: usize) -> Result<()>;
+
+    /// Records that the patch with number patch_number failed to boot, and ensures
+    /// that it will never be returned as the next boot or last booted patch.
+    fn record_boot_failure_for_patch(&mut self, patch_number: usize) -> Result<()>;
```
eseidel, line 63: I wonder if it's fragile that you can call "success" any number of times but if you ever "fail" we roll back.  I'm not sure what options we have of course if a "failure" ever happens.  Currently the only way things can fail is if Dart fails to initialize, which feels like a pretty severe, deserving-of-rollback case.  But we may have other gradations of "failure" in the future 🤷  (Not sure what we'd do with other levels of failure though.)

library/src/cache/patch_manager.rs
eseidel: I'm pretty sure the previous version wasn't necessarily on disk?  I believe I added this because whenever a patch failed before we'd keep trying to download it again?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,683 @@
...
+
+    /// Resets the patch manager to its initial state, removing all patches. This is
+    /// intended to be used when a new release version is installed.
+    fn reset(&mut self) -> Result<()>;
+}
+
+// This allows us to use the Debug trait on dyn ManagePatches, which is
+// required to have it as a property of UpdaterState.
+impl Debug for dyn ManagePatches {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
```
eseidel, line 77: CodeCov would like you to cover this, which should be trivial. 🤣

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,683 @@
...
+
+        Self {
+            root_dir,
+            patches_state,
+        }
+    }
+
+    fn load_patches_state(root_dir: &Path) -> Option<PatchesState> {
+        let path = root_dir.join(PATCHES_STATE_FILE_NAME);
+        disk_io::read(&path).ok()
```
eseidel, line 113: Do we need to log the failure at all?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,683 @@
...
+
+    /// The directory where artifacts for the patch with the given number are stored.
+    fn dir_for_patch_number(&self, patch_number: usize) -> PathBuf {
+        self.patch_artifacts_dir().join(patch_number.to_string())
+    }
+
+    /// The path to the runnable patch artifact with the given number. Runnable patch artifact files are
+    /// named <patch_number>.vmcode
+    fn patch_artifact_path(&self, patch_number: usize) -> PathBuf {
+        self.dir_for_patch_number(patch_number).join("dlc.vmcode")
```
eseidel, line 134: Should dlc.vmcode be a constant somewhere?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,683 @@
...
+
+    /// Checks that the patch with the given number:
+    ///   - Has an artifact on disk
+    ///   - That artifact on disk is the same size it was when it was installed
+    ///
+    /// Returns Ok if the patch is bootable, or an error if it is not.
+    fn validate_patch_is_bootable(&self, patch: &PatchMetadata) -> Result<()> {
+        let artifact_path = self.patch_artifact_path(patch.number);
+        if !Path::exists(&artifact_path) {
+            bail!(
```
eseidel, line 152: codecov would like you to cover this, should be pretty easy?

eseidel (commented):
I haven't read it all yet.

library/src/cache/disk_io.rs
bryanoltman: 😬  correct, fixed

library/src/cache/disk_io.rs
bryanoltman: I discovered `with_context` about halfway through writing this, thought I caught all of these. Fixed!

library/src/cache/disk_io.rs
bryanoltman: I thought `?` seemed a bit cleaner than having `.unwrap()` everywhere. See https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#tests-and-

library/src/cache/patch_manager.rs
bryanoltman: The updater's directory structure now looks like:

```
shorebird_updater/
  state.json
  patches_state.json
  patches/
    1/
       dlc.vmcode
    2/
       dlc.vmcode
```

Is that helpful?

library/src/cache/disk_io.rs
eseidel: That's AWESOME.

library/src/cache/patch_manager.rs
bryanoltman: It generally does not. Removed.

library/src/cache/patch_manager.rs
bryanoltman: Added an error log

library/src/cache/patch_manager.rs
eseidel: Which part of that does PatchMetadata refer to?

eseidel: And will you have to change it when the next patch you make to this Rust code will be to support multiple vmcode files for iOS? 🤣

eseidel: This code makes me wish that .dmg files (or some other seamless container, maybe .tar?) worked well on iOS. So that you could have multiple files appear as one file on disk.

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+    /// Metadata about the patches we have downloaded that is persisted to disk.
+    patches_state: PatchesState,
+}
+
+impl PatchManager {
+    /// Creates a new PatchManager with the given root directory. This directory is
+    /// assumed to exist. The PatchManager will use this directory to store its
+    /// state and patch binaries.
+    pub fn with_root_dir(root_dir: PathBuf) -> Self {
+        let patches_state = Self::load_patches_state(&root_dir).unwrap_or_default();
```
eseidel, line 106: Does the caller care that this might silently make a new one on failure?

eseidel: I guess it's not silent, you get an error! below.

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+        if !Path::exists(&artifact_path) {
+            bail!(
+                "Patch {} does not exist at {}",
+                patch.number,
+                artifact_path.display()
+            );
+        }
+
+        let artifact_size_on_disk = std::fs::metadata(&artifact_path)?.len();
+        if artifact_size_on_disk != patch.size {
```
eseidel, line 173: Makes me wonder if we should report the boot size as part of our patch_available callback to the server, or some other mechanism for possibly detecting tampering in the field.

eseidel: (Not something to do in this patch, more just seeding your thoughts.)

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+        std::fs::remove_dir_all(&patch_dir)
+            .with_context(|| format!("Failed to delete patch dir {}", &patch_dir.display()))
+    }
+
+    /// Attempts to use the last successfully booted patch as the next boot patch. If the last successfully
+    /// booted patch is not bootable or has the same number as the patch we're falling back from, we clear it.
+    fn try_fall_back_to_last_booted_patch(&mut self) {
+        if let Some(next_boot_patch) = self.patches_state.next_boot_patch {
+            // If we have a next_boot_patch that we're falling back from, delete its artifacts.
+            if let Err(e) = self.delete_patch_artifacts(next_boot_patch.number) {
```
eseidel, line 199: I guess I didn't expect this "cleanup the current patch" to be in this method. 🤷

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+                        last_boot_patch.number, e
+                    );
+                }
+            }
+        }
+    }
+}
+
+impl ManagePatches for PatchManager {
+    fn add_patch(&mut self, patch_number: usize, file_path: &Path) -> Result<()> {
```
eseidel, line 235: This appears to mean a patch as just a single dlc.vmcode file.  (which is fine, you just may be refactoring this soon). :)

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+    fn next_boot_patch(&mut self) -> Option<PatchInfo> {
+        let next_boot_patch = match self.patches_state.next_boot_patch {
+            Some(patch) => patch,
+            None => return None,
+        };
+
+        if let Err(e) = self.validate_patch_is_bootable(&next_boot_patch) {
+            error!("Patch {} is not bootable: {}", next_boot_patch.number, e);
+
+            self.try_fall_back_to_last_booted_patch();
```
eseidel, line 292: Does this need to report back that it was successful/not in falling back? I guess it's job is to leave us in a consistent state of things bootable?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+            .map(|patch| self.patch_info_for_number(patch.number))
+    }
+
+    fn record_boot_success_for_patch(&mut self, patch_number: usize) -> Result<()> {
+        let next_boot_patch = self
+            .patches_state
+            .next_boot_patch
+            .context("No next_boot_patch")?;
+
+        if next_boot_patch.number != patch_number {
```
eseidel, line 312: Is next_boot_patch the current patch we're booting from?  Won't this fail if the updater thread somehow races ahead of dart initialization and has already installed a new patch before this is called?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+            bail!(
+                "Attempted to record boot success for patch {} but next_boot_patch is {}",
+                patch_number,
+                next_boot_patch.number
+            );
+        }
+
+        if let Some(current_patch) = self.patches_state.last_booted_patch {
```
eseidel, line 320: So a side-effect of successfully booting is that we clean up previously booted patches?

eseidel: Couldn't this end up cleaning up the "next_boot_patch" if they're the same?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+            self.delete_patch_artifacts(current_patch.number)?;
+        }
+
+        self.patches_state.last_booted_patch = Some(next_boot_patch);
+        self.save_patches_state()
+    }
+
+    fn record_boot_failure_for_patch(&mut self, patch_number: usize) -> Result<()> {
+        let next_boot_patch = self
```
eseidel, line 329: I find it confusing to read "next boot" when talking about the patch we're currently running from.

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+        assert!(manager.add_patch(1, file_path).is_ok());
+        assert_eq!(manager.highest_seen_patch_number(), Some(1));
+
+        // Add patch 4, expect 4 to be the highest patch number we've seen
+        let file_path = &temp_dir.path().join("patch.vmcode");
+        std::fs::write(file_path, patch_file_contents)?;
+        assert!(manager.add_patch(4, file_path).is_ok());
+        assert_eq!(manager.highest_seen_patch_number(), Some(4));
+
+        // Add patch 3, expect 4 to still be the highest patch number we've seen
```
eseidel, line 464: Does this imply that the way we'll end up implementing rollback is by copying a patch to a new patch number?

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+mod get_next_boot_patch_tests {
+    use super::*;
+    use anyhow::Result;
+    use tempdir::TempDir;
+
+    #[test]
+    fn returns_none_if_no_next_boot_patch() {
+        let temp_dir = TempDir::new("patch_manager").unwrap();
+        let mut manager = PatchManager::with_root_dir(temp_dir.path().to_owned());
+        assert!(manager.next_boot_patch().is_none());
```
eseidel, line 516: I'm not sure if this or assert_eq!(value, None) is more idiomatic (you seem to use both in this file).

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,883 @@
...
+        assert!(manager.add_patch(1, file_path).is_ok());
+        assert!(manager.record_boot_success_for_patch(1).is_ok());
+
+        // Add patch 2, pretend it failed to boot.
+        let file_path = &temp_dir.path().join("patch2.vmcode");
+        std::fs::write(file_path, patch_file_contents)?;
+        assert!(manager.add_patch(2, file_path).is_ok());
+        assert!(manager.record_boot_failure_for_patch(2).is_ok());
+
+        // Write junk to patch 1's artifact. This should prevent us from falling back to it.
```
eseidel, line 609: I love these tests, btw.

library/src/updater.rs
```diff
@@ -352,7 +354,7 @@ where
 /// This may be changed any time `update()` or `start_update_thread()` are called.
 pub fn next_boot_patch() -> anyhow::Result<Option<PatchInfo>> {
     with_config(|config| {
-        let state =
+        let mut state =
```
eseidel, line 357: Why is this mut?  You might want to use some tricks to internal mutability in next_boot_patch() if it's mutating (I can look up how I did it in safe_wren).

eseidel (commented):
My only real concern is if there is a race re: next_boot_patch between the updater thread and the success callback.  It's not clear to me (at least from the naming) that the succes/failure callbacks are using the correct patch in their checks.

library/src/cache/patch_manager.rs
bryanoltman: I imagine asset support will probably bring a few other changes as well :)

library/src/updater.rs
bryanoltman: Agreed, will update

library/src/cache/patch_manager.rs
bryanoltman: Ah, I understand now. `PatchMetadata` comes baked with the assumption that the patch is a single file. We'll need to change this to a map of file names to sizes (or a collection of a yet-to-be-defined `PatchFile` type) down the road.

library/src/cache/patch_manager.rs
bryanoltman: Done

library/src/cache/patch_manager.rs
bryanoltman: It wasn't originally, but this function is used as a "we couldn't boot from the patch we wanted to, see if we can boot from the last good one" operation, and cleaning up the one we're falling back from is a part of that. It could maybe have a better name, but I'm not sure what that would be atm

library/src/cache/patch_manager.rs
bryanoltman: Yeah, there's probably a better name to communicate "the patch we're going to boot from when we next launch, that we maybe have or haven't booted from before". Maybe just `boot_patch`?

library/src/cache/patch_manager.rs
bryanoltman: Yeah, we print the error, and there's not much we can do in the way of recovery if we can't deserialize the json file

library/src/cache/disk_io.rs
bryanoltman: This file has 100% coverage according to the https://app.codecov.io/gh/shorebirdtech/updater/pull/93/blob/library/src/cache/disk_io.rs, resolving

library/src/cache/patch_manager.rs
bryanoltman: Done

library/src/cache/patch_manager.rs
eseidel: Yeah, assuming I'm reading this correctly, this method is such that we only get to boot once from a patch and then it deletes the patch 🤣   Am I reading that wrong?

bryanoltman: Nope, you're reading correctly. We just happened to only call this function for patches that had not previously booted (probably incorrectly). Fixed!

library/src/cache/patch_manager.rs
bryanoltman: I'm not sure what specifically you mean by rollback. Are you referring to the case where a new patch fails to boot, or a developer-initiated rollback feature that we'll build in the future?

library/src/cache/patch_manager.rs
bryanoltman: I'm not able to find a clear answer after a couple minutes of googling, but I personally prefer `is_none()` (in the same way I'd prefer `assert!(my_string.is_empty())` to `assert_eq!(my_string, "")`). Just pushed a change to standardize.

library/src/updater.rs
bryanoltman: This updater state is mutable because it calls `self.patch_manager.next_boot_patch()`, which can mutate the patch manager that it owns. A little weird, I agree.

library/src/cache/patch_manager.rs
eseidel: I was imagining how a future feature might work and realizing that clients are going to store state about what patch they've seen and will refuse to go backwards to a patch.  So if we implement rollback we'll need to "copy" the patch and assign it a new number.

library/src/updater.rs
eseidel: I believe rust has ways to hide internal mutability through a Ref, but I'd have to look at my old code.

eseidel: RefCell is what I'm thinking of.

library/src/cache/patch_manager.rs
bryanoltman: Just to make sure I understand (and to spell it out), the scenario we're worried about is:

1. The app launches using no patch.
2. The app downloads a new patch (call it `patch 1`, stores it as the `next_boot_patch`.
3. A new patch is published (call it `patch 2`).
4. The app restarts, booting from `patch 1` and downloading `patch 2`
5. `patch 2` finishes downloading and is marked `next_boot_patch` before `patch 1` finishes booting.
6. `record_boot_success_for_patch` is called for `patch 1`, but this method doesn't record it because `patch 2` is the next boot patch.

Is that correct?

library/src/cache/patch_manager.rs
eseidel: 🤷  You have a nice comment about it that I just didn't see the first time I think.

library/src/cache/patch_manager.rs
bryanoltman: I think we generally need some mechanism of reporting that we decided a patch wasn't bootable. Noodling on that now

library/src/cache/patch_manager.rs
bryanoltman: >  if we implement rollback we'll need to "copy" the patch and assign it a new number.

Yeah, this is how I imagined it would work – take the artifacts from our rollback target patch and make it patch N+1

library/src/updater.rs
bryanoltman: Ah gotcha. It looks like RefCell is [unsafe](https://doc.rust-lang.org/book/ch15-05-interior-mutability.html), so I'm inclined to stick to the odd-looking `mut state` for now unless you have strong objections.

eseidel: sgtm.

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,908 @@
...
+                error!("Failed to delete patch dir {}: {}", patch_dir.display(), e);
+                e
+            })
+            .with_context(|| format!("Failed to delete patch dir {}", &patch_dir.display()))
+    }
+
+    /// Attempts to use the last successfully booted patch as the next boot patch. If the last successfully
+    /// booted patch is not bootable or has the same number as the patch we're falling back from, we clear it.
+    fn try_fall_back_to_last_booted_patch(&mut self) {
+        if let Some(next_boot_patch) = self.patches_state.next_boot_patch {
```
eseidel, line 201: This feels like this is "given this patch number, please remove it from your patches_state".

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,908 @@
...
+
+            if let Some(last_boot_patch) = self.patches_state.last_booted_patch {
+                if last_boot_patch.number == next_boot_patch.number {
+                    // If the last booted patch is the same as the next boot patch, clear it.
+                    self.patches_state.last_booted_patch = None;
+                }
+            }
+        }
+
+        if let Some(last_boot_patch) = self.patches_state.last_booted_patch {
```
eseidel, line 214: And this is, "if we still have a last_booted_patch, see if it's still validate an re-install it".

library/src/cache/patch_manager.rs
```diff
@@ -0,0 +1,908 @@
...
+            if self.validate_patch_is_bootable(&last_boot_patch).is_ok() {
+                // If we think we can still boot from the last booted patch, set it as the next_boot_patch.
+                self.patches_state.next_boot_patch = Some(last_boot_patch);
+            } else {
+                self.patches_state.last_booted_patch = None;
+                let _ = self.delete_patch_artifacts(last_boot_patch.number);
```
eseidel, line 220: I presume this validation of the next_boot_patch happens elsewhere?  Is this just a copy of some of the install logic?

library/src/cache/patch_manager.rs
bryanoltman: Not sure I follow

bryanoltman: Ah, I just needed to continue reading. Your understanding is correct. This function is "remove the next boot patch if we have one (and the last booted patch if they're the same), and provide the last booted patch if it exists and is valid"

library/src/cache/patch_manager.rs
bryanoltman: Exactly

library/src/cache/patch_manager.rs
eseidel: Correct.

library/src/updater.rs
```diff
@@ -411,51 +414,57 @@ pub fn report_launch_failure() -> anyhow::Result<()> {
...
-        } else {
-            Ok(())
+            // We didn't boot from a patch, so there's nothing to do.
+            None => return Ok(()),
+        };
+
+        let maybe_previous_boot_patch = state.current_boot_patch();
+
+        state.record_boot_success_for_patch(next_boot_patch.number)?;
+
```
eseidel, line 438: One could use state.known_good_patch_list() which just returned either [] or [current_boot_patch] and then this code can just check known_good_patch_list().contains(current_boot_patch.number)?

🤷

eseidel: state.is_known_good_patch(number)?

eseidel (approved):

PR #111, opened 2024-01-08
feat: Update C API to consume Read+Seek callbacks

bryanoltman:
## Description

Updates the update function to accept an object that implements Read+Seek. This will allow the flutter engine to provide the base snapshot data to the updater directly and allow us to move away from having the updater find and read base snapshot files.

Still WIP, needs cleanup and test fixing.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


library/src/c_api.rs
```diff
@@ -46,7 +47,65 @@ pub struct AppParameters {
...
+impl Seek for CFile {
+    fn seek(&mut self, pos: std::io::SeekFrom) -> std::io::Result<u64> {
+        let (offset, whence) = match pos {
+            std::io::SeekFrom::Start(offset) => (offset as i64, libc::SEEK_SET),
+            std::io::SeekFrom::End(offset) => (offset, libc::SEEK_END),
+            std::io::SeekFrom::Current(offset) => (offset, libc::SEEK_CUR),
+        };
+        let result = (self.file_callbacks.seek)(self.handle, offset, whence);
+        if result < 0 {
+            Err(std::io::Error::last_os_error())
```
eseidel, line 101: Why last_os_error?

library/src/updater.rs
```diff
@@ -136,6 +84,21 @@ pub struct AppConfig {
     pub original_libapp_paths: Vec<String>,
 }
 
+// Struct which provides file callbacks in a rust-like way
+// Has two callbacks.  One open, which returns a Read+Seek object
+// The other a close which takes the Read+Seek object
+
+pub trait ReadSeek: Read + Seek {}
+
+/// TODO
```
eseidel, line 93: todo

library/src/updater.rs
```diff
@@ -231,33 +194,19 @@ fn check_hash(path: &Path, expected_string: &str) -> anyhow::Result<()> {
...
+
+#[cfg(any(target_os = "android", test))]
+fn patch_base(config: &UpdateConfig) -> anyhow::Result<Box<dyn ReadSeek>> {
+    let base_r = crate::android::open_base_lib(&config.libapp_path, "libapp.so")?;
+    Ok(Box::new(base_r))
+}
+
+#[cfg(not(any(target_os = "android", test)))]
+fn patch_base(config: &UpdateConfig) -> anyhow::Result<Box<dyn ReadSeek>> {
+    // TODO use a nonempty string here
```
eseidel, line 207: Todo

library/src/c_api.rs
bryanoltman: Copilot :) Updated.

library/src/c_api/mod.rs
```diff
@@ -46,6 +50,15 @@ pub struct AppParameters {
     pub code_cache_dir: *const libc::c_char,
 }
 
+#[derive(Clone, Copy, Debug)]
+#[repr(C)]
+pub struct FileCallbacks {
+    pub open: extern "C" fn() -> *mut libc::c_void,
```
eseidel, line 56: I'd give them argument names and comments for your own later sanity, but that can also be a follow-up.

library/src/c_api/c_file.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use serial_test::serial;
+
+    use super::*;
+
+    static mut OPEN_CALL_COUNT: usize = 0;
```
eseidel, line 71: I guess I'm confused why these are static, instead of held off the handle pointer?  Doesn't matter a ton, but this will cause you to have to run tests in serial rather than letting them run parallel as they do by default.

library/src/c_api/c_file.rs
```diff
@@ -0,0 +1,228 @@
...
+            READ_ARGS.clear();
+            SEEK_ARGS.clear();
+        }
+    }
+
+    extern "C" fn fake_open() -> *mut libc::c_void {
+        unsafe {
+            OPEN_CALL_COUNT += 1;
+        }
+        std::ptr::null_mut()
```
eseidel, line 90: I would think that you'd want your rust wrapper to detect a null handle return and fail. :)

library/src/c_api/mod.rs
```diff
@@ -107,13 +120,17 @@ where
 #[no_mangle]
 pub extern "C" fn shorebird_init(
     c_params: *const AppParameters,
+    c_file_callbacks: FileCallbacks,
     c_yaml: *const libc::c_char,
```
eseidel, line 124: Soon we can move the reading of the shorebird.yaml into the FileCallbacks if we want. 🤣

library/src/updater.rs
```diff
@@ -103,7 +115,11 @@ fn libapp_path_from_settings(original_libapp_paths: &[String]) -> Result<PathBuf
 /// The yaml string is the contents of the `shorebird.yaml` file.
 /// The `AppConfig` struct is information about the running app and where
 /// the updater should keep its cache.
-pub fn init(app_config: AppConfig, yaml: &str) -> Result<(), UpdateError> {
+pub fn init(
+    app_config: AppConfig,
+    file_provider: Box<dyn ExternalFileProvider>,
```
eseidel, line 120: I do think you're going to end up with this being Arc in a later patch, since that makes the lifetime more clear.

eseidel (approved):
There is definitely cleanup we can do, but I think we can also just land and iterate.

library/src/c_api/c_file.rs
```diff
@@ -0,0 +1,228 @@
...
+        unsafe {
+            OPEN_CALL_COUNT = 0;
+            CLOSE_CALL_COUNT = 0;
+            READ_ARGS.clear();
+            SEEK_ARGS.clear();
+        }
+    }
+
+    extern "C" fn fake_open() -> *mut libc::c_void {
+        unsafe {
```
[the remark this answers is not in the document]
bryanoltman: ^ without the unsafe block

library/src/c_api/c_file.rs
bryanoltman: ¯\_(ツ)_/¯ this is just test code

library/src/c_api/c_file.rs
bryanoltman: The `FileCallbacks` functions are static, and I'm not sure there's a way to associate a `self` with them. Happy to be wrong on this one

library/src/c_api/c_file.rs
bryanoltman: Misread this initially, good call and will update

library/src/c_api/c_file.rs
eseidel: Yeah, it would need to be held off the handle.  You're correct, it's just test code and it's fine as is.

library/src/c_api/c_file.rs
bryanoltman: Done.

library/src/c_api/mod.rs
bryanoltman: Done

PR #113, opened 2024-01-12
Refactor posix file i/o to c_api

bryanoltman:
WIP, need to fix tests and make sure this actually works


eseidel (commented):
Looks good to me!

PR #138, opened 2024-04-18
feat: update patch manager to only attempt to boot from a patch once

bryanoltman:
## Description

Draft while I test

Fixes https://github.com/shorebirdtech/engine/issues/79

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


library/src/cache/patch_manager.rs
```diff
@@ -290,6 +337,24 @@ impl ManagePatches for PatchManager {
...
+            .next_boot_patch
+            .context("No next_boot_patch")?;
+
+        if next_boot_patch.number != patch_number {
+            bail!(
+                "Attempted to record boot success for patch {} but next_boot_patch is {}",
+                patch_number,
+                next_boot_patch.number
+            );
+        }
```
bryanoltman, line 352: Not covered, but my next change is going to be a refactor to remove the patch number argument, so this is somewhat irrelevant.

library/src/cache/patch_manager.rs
```diff
@@ -316,16 +381,16 @@ impl ManagePatches for PatchManager {
...
 
-        if next_boot_patch.number != patch_number {
+        if last_attempted_patch.number != patch_number {
             bail!(
                 "Attempted to record boot failure for patch {} but should have booted from {}",
                 patch_number,
-                next_boot_patch.number
+                last_attempted_patch.number
             );
         }
```
bryanoltman, line 395: Same here re: the refactor to remove the patch number argument

library/README.md
```diff
@@ -132,26 +132,17 @@ NDK_HOME=$HOME/Documents/GitHub/engine/src/third_party/android_tools/ndk cargo n
...
-
-- Patches are cached on disk in "slots".
-- There is a currently active slot (the one that is booted).
-- Patches are identified by base revision + patch number.
-- A given slot is:
-  - `empty`: No update is installed.
-  - `pending`: An update is installed but has not been validated.
-  - `valid`: An update is installed and has been validated.
-- Validation is a temporary state. Patches/slots are revalidated on boot.
-
```
bryanoltman, line 154: Removed this as it's no longer accurate (we no longer use the slot system)

library/src/cache/patch_manager.rs
```diff
@@ -52,11 +55,19 @@ pub trait ManagePatches {
     /// or None if no patch is installed.
     fn last_successfully_booted_patch(&self) -> Option<PatchInfo>;
 
+    /// The patch we most recently attempted to boot. Will be the same as
+    /// last_successfully_booted_patch if the last boot was successful.
+    fn last_attempted_boot_patch(&self) -> Option<PatchInfo>;
```
eseidel, line 60: Looks like PatchInfo and PatchMetadata are the same thing, just with internal/external types? https://github.com/shorebirdtech/updater/blob/0a0cd183f529cb9f2bab4d8ef63c893dd9f3a684/library/src/cache/mod.rs#L9

library/src/cache/patch_manager.rs
```diff
@@ -179,6 +190,37 @@ impl PatchManager {
             );
         }
 
+        // Verify that this patch isn't known bad.
+        // Have we booted from any patch before?
```
eseidel, line 194:
```suggestion
        // Have we attempted to boot from any patch before?
```

library/src/cache/patch_manager.rs
```diff
@@ -179,6 +190,37 @@ impl PatchManager {
...
+        if let Some(last_attempted_patch) = self.patches_state.last_attempted_patch {
+            // This would ideally be part of the if statement above, but
+            // Rust does not like that, at least for the time being.
+            // error[E0658]: `let` expressions in this position are unstable
+            // https://github.com/rust-lang/rust/issues/53667
+
+            // Was it this patch?
```
eseidel, line 201:
```suggestion
            // If the last patch we tried to boot from was this patch, that's OK so long as
            // it was successful.
```

eseidel (changes requested):

library/src/cache/patch_manager.rs
```diff
@@ -32,6 +32,9 @@ struct PatchesState {
     /// The patch we are currently running, if any.
     last_booted_patch: Option<PatchMetadata>,
 
+    /// The last patch we attempted to boot, if any.
+    last_attempted_patch: Option<PatchMetadata>,
```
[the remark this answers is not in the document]
bryanoltman: > Does this mean we'll potentially hold lots of copies of a given patch metadata? Will that be a problem?

We'll at most hold three, and given that these are just two 64-bit fields, I don't think there's much to be concerned about.

> I'm not sure why we hold onto the size as part of this?

It's our very primitive "was this corrupted" check – if the size has changed from what we recorded, we don't attempt to boot from the patch.

library/src/cache/patch_manager.rs
bryanoltman: Not quite. The naming could certainly be better, but PatchInfo has a number and a path, PatchMetadata has a number and a size. PatchMetadata is internal to PatchManager, PatchInfo is the interface used by other modules to talk to PatchManager. We could probably consolidate these, but that would probably be better as a future refactor.

library/src/cache/patch_manager.rs
```diff
@@ -179,6 +190,37 @@ impl PatchManager {
             );
         }
 
+        // Verify that this patch isn't known bad.
+        // Have we booted from any patch before?
+        if let Some(last_attempted_patch) = self.patches_state.last_attempted_patch {
+            // This would ideally be part of the if statement above, but
+            // Rust does not like that, at least for the time being.
+            // error[E0658]: `let` expressions in this position are unstable
+            // https://github.com/rust-lang/rust/issues/53667
```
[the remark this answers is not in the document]
bryanoltman: Updated the comment for clarity.

library/src/cache/patch_manager.rs
bryanoltman: Addressed

library/src/cache/patch_manager.rs
bryanoltman: Addressed

library/src/cache/patch_manager.rs
```diff
@@ -179,6 +190,37 @@ impl PatchManager {
...
+
+            // Was it this patch?
+            if last_attempted_patch.number == patch.number {
+                // Have we successfully booted from any patch before?
+                if let Some(last_successful_patch) = self.patches_state.last_booted_patch {
+                    // Was it this patch?
+                    if last_attempted_patch.number != last_successful_patch.number {
```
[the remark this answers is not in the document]
bryanoltman: I _believe_ these should all be covered. I've refactored slightly for better readability, so this isn't quite the nest of `if` statements that it was before.

library/src/cache/patch_manager.rs
```diff
@@ -179,9 +190,52 @@ impl PatchManager {
...
+        }
+
         Ok(())
     }
 
+    /// Whether the given patch number is the last one we attempted to boot
+    /// (whether it was successful or not).
+    fn is_patch_last_attempted_patch(&self, patch_number: usize) -> bool {
+        self.patches_state
+            .last_attempted_patch
```
eseidel, line 227: Would this also be self.patches_state.last_attempted_patch == Some(patch_number)?

eseidel: Oh, I see, last_attempted_patch isn't itself a number, it's a number + size.

library/src/cache/patch_manager.rs
```diff
@@ -179,9 +190,52 @@ impl PatchManager {
             );
         }
 
+        // Verify that we haven't tried and failed to boot this patch.
+        if self.is_patch_last_attempted_patch(patch.number) {
+            match self.last_successful_boot_patch_number() {
```
eseidel, line 195: last_successful_boot_patch_number still feels like the wrong check.  We care if this patch is a known good patch or not (have we *ever* successfully booted from it), not if the last patch we happened to boot from was this one.

eseidel (commented):
Maybe we should discuss on vc.

library/src/cache/patch_manager.rs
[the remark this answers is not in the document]
bryanoltman: Numbering these:

1. If this is a known-bad patch (already covered), don't let it boot, period.
2. Is this already a known good patch, if so, let it boot, period.
3. If this isn't already a known good patch, was our last attempt an attempt to boot from this? If so, fail.

We get (with some comments removed for clarity)

```rust
        if self.is_patch_last_attempted_patch(patch.number) { // Case 3
            match self.last_successful_boot_patch_number() {
                Some(last_successful_patch_number)
                    if last_successful_patch_number == patch.number =>
                {
                    // Case 2 
                    // We have successfully booted from this patch before.
                }
                _ => {
                    // Case 1
                    bail!(
                        "Already attempted and failed to boot patch {}",
                        patch.number
                    )
                }
            }
        }
```

I cleaned this up to make it less of a nested `if` readability headache.

library/src/cache/patch_manager.rs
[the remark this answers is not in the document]
eseidel: Yeah, I'm not at all concerned about memory usage, more about correctness.  I don't think I have a real concern, here, but in general whenever we're storing the same data in multiple places, it invites possibilities of having diverging copies of the data.

library/src/cache/patch_manager.rs
eseidel: This was my confusion thinking we still had a known_good and known_bad list.  In our current updater whenever we record a new "successful" patch we forget all knowledge about past patches good or bad state.

library/src/cache/patch_manager.rs
```diff
@@ -179,9 +190,46 @@ impl PatchManager {
...
 
+        // Verify that we haven't tried and failed to boot this patch.
+        if self.is_patch_last_attempted_patch(patch.number) {
+            // We are trying to boot from the same patch that we tried to boot from last time.
+
+            match self.last_successful_boot_patch_number() {
+                Some(last_successful_patch_number)
+                    if last_successful_patch_number == patch.number =>
+                {
+                    // We have successfully booted from this patch before.
```
eseidel, line 201:
```suggestion
                    // Our last boot attempt was this patch, and we've successfully booted from this
                    // patch before.  This patch is safe to boot from.
```

library/src/cache/patch_manager.rs
```diff
@@ -179,9 +190,46 @@ impl PatchManager {
             );
         }
 
+        // Verify that we haven't tried and failed to boot this patch.
```
eseidel, line 193:
```suggestion
        // If the last boot we tried was this patch, make sure we succeeded or the patch is bad.
```

eseidel (approved):

PR #142, opened 2024-04-19
chore: remove unused package

eseidel:
Much more to remove here, but this is a start.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1018,7 +1018,7 @@ dependencies = [
 
 [[package]]
 name = "patch"
-version = "0.1.0"
+version = "0.2.1"
 dependencies = [
  "bidiff",
  "comde",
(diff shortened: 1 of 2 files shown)


eseidel:
I don't understand the "patch" version change.

eseidel:
In cargo.lock

Cargo.lock
```diff
@@ -1018,7 +1018,7 @@ dependencies = [
 
 [[package]]
 name = "patch"
-version = "0.1.0"
+version = "0.2.1"
```
eseidel, line 1021: I don't understand this change.

eseidel: But our "patch" library is version = "0.2.1" 🤷

bryanoltman:
Patch is version 0.2.1 (see [here](https://github.com/shorebirdtech/updater/blob/main/patch/Cargo.toml#L3)). I believe it's just updating to reflect that.

bryanoltman (approved):

PR #149, opened 2024-04-22
fix: verify failed patch is next_boot_patch before clearing

bryanoltman:
## Description

Removes assumption that a failed patch is `next_boot_patch` in fallback behavior.

Fixes https://github.com/shorebirdtech/updater/issues/148

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -837,6 +837,7 @@ mod fall_back_tests {
         manager.add_patch_for_test(&temp_dir, 1)?;
 
         // Start booting from patch 1.
+        manager.record_boot_start_for_patch(1)?;
 
         // Download patch 2 before patch 1 finishes booting.
         manager.add_patch_for_test(&temp_dir, 2)?;
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -248,6 +248,7 @@ impl PatchManager {
     /// successfully booted patch. If the last successfully booted patch is not bootable or has the same number
     /// as the patch we're falling back from, we clear it as well.
     fn try_fall_back_from_patch(&mut self, bad_patch_number: usize) {
+        // No need to log failure – delete_patch_artifacts logs for us.
         let _ = self.delete_patch_artifacts(bad_patch_number);
 
         if let Some(next_boot_patch) = self.patches_state.next_boot_patch {
@@ -271,6 +272,7 @@ impl PatchManager {
                 self.patches_state.next_boot_patch = Some(last_boot_patch);
             } else {
                 self.patches_state.last_booted_patch = None;
+                // No need to log failure – delete_patch_artifacts logs for us.
                 let _ = self.delete_patch_artifacts(last_boot_patch.number);
             }
         }
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -195,25 +195,30 @@ impl PatchManager {
             .with_context(|| format!("Failed to delete patch dir {}", &patch_dir.display()))
     }
 
-    /// Attempts to use the last successfully booted patch as the next boot patch. If the last successfully
-    /// booted patch is not bootable or has the same number as the patch we're falling back from, we clear it.
-    fn try_fall_back_to_last_booted_patch(&mut self) {
-        if let Some(next_boot_patch) = self.patches_state.next_boot_patch {
-            // If we have a next_boot_patch that we're falling back from, delete its artifacts.
-            let _ = self.delete_patch_artifacts(next_boot_patch.number);
-            self.patches_state.next_boot_patch = None;
+    /// Deletes artifacts for the provided bad_patch_number and attempts to set the next_boot_patch to the last
+    /// successfully booted patch. If the last successfully booted patch is not bootable or has the same number
+    /// as the patch we're falling back from, we clear it as well.
+    fn try_fall_back_from_patch(&mut self, bad_patch_number: usize) {
+        let _ = self.delete_patch_artifacts(bad_patch_number);
 
-            if let Some(last_boot_patch) = self.patches_state.last_booted_patch {
-                if last_boot_patch.number == next_boot_patch.number {
-                    // If the last booted patch is the same as the next boot patch, clear it.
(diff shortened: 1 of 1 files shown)


library/src/cache/patch_manager.rs
```diff
@@ -195,25 +195,30 @@ impl PatchManager {
...
     }
 
-    /// Attempts to use the last successfully booted patch as the next boot patch. If the last successfully
-    /// booted patch is not bootable or has the same number as the patch we're falling back from, we clear it.
-    fn try_fall_back_to_last_booted_patch(&mut self) {
+    /// Deletes artifacts for the provided bad_patch_number and attempts to set the next_boot_patch to the last
+    /// successfully booted patch. If the last successfully booted patch is not bootable or has the same number
+    /// as the patch we're falling back from, we clear it as well.
+    fn try_fall_back_from_patch(&mut self, bad_patch_number: usize) {
+        let _ = self.delete_patch_artifacts(bad_patch_number);
```
eseidel, line 202: It appears you're intentionally swallowing Err here?  Are you sure that's what you want?

bryanoltman: `delete_patch_artifacts` logs the error – we don't need to do anything with it here. Failing to clear bad patch artifacts shouldn't stop us from marking the patch as bad and verifying that we can roll back (or not)

eseidel: I think you'll want to leave your future self that as a comment. :)

eseidel: Caused me to re-read https://stackoverflow.com/questions/51141672/how-do-i-ignore-an-error-returned-from-a-rust-function-and-proceed-regardless just now. :)

bryanoltman: > I think you'll want to leave your future self that as a comment. :)

Done

library/src/cache/patch_manager.rs
```diff
@@ -244,28 +244,35 @@ impl PatchManager {
...
-
-            if let Some(last_boot_patch) = self.patches_state.last_booted_patch {
-                if last_boot_patch.number == next_boot_patch.number {
-                    // If the last booted patch is the same as the next boot patch, clear it.
-                    self.patches_state.last_booted_patch = None;
+            // If our next boot patch is bad_patch_number, clear it.
+            if next_boot_patch.number == bad_patch_number {
+                self.patches_state.next_boot_patch = None;
+
+                // If the last booted patch is the same as the next boot patch, clear it.
```
eseidel, line 259: This comment is wrong?  Also, why is this inside the if?  Should last_booted_patch be cleared anytime it's the same as bad_patch_number?  Also, why would we ever mark a last_booted_patch as bad?

eseidel: I think we shouldn't ever mark a last_booted_patch as bad, right?  Feels like something we also need a test for.

bryanoltman: This could be cleaned up a bit, doing that now.

> I think we shouldn't ever mark a last_booted_patch as bad, right? Feels like something we also need a test for.

If we have successfully booted from a patch whose size changes on disk (our proxy for file corruption), it has become bad and we don't want to try to boot from it again. We have a test for this: `clears_next_and_last_patches_if_both_fail_validation`. Adding comments and cleaning up now.

bryanoltman: Addressed

library/src/cache/patch_manager.rs
```diff
@@ -244,28 +244,30 @@ impl PatchManager {
...
+                self.patches_state.next_boot_patch = None;
             }
         }
 
+        // If we think we can still boot from the last booted patch, set it as the next_boot_patch.
+        // If something happened to render the last boot patch unbootable, clear it and delete its artifacts.
         if let Some(last_boot_patch) = self.patches_state.last_booted_patch {
-            if self.validate_patch_is_bootable(&last_boot_patch).is_ok() {
-                // If we think we can still boot from the last booted patch, set it as the next_boot_patch.
+            if last_boot_patch.number != bad_patch_number
```
eseidel, line 264: I'm still not sure why bad_patch_number ever matters to a last_boot_patch?

If we're trying to catch the case where:
- success boot patch 1
- install patch 2
- fail to boot patch 2
- fail to boot patch 1
then lets explicitly test that.

Previously I was under the assumption that:

- success boot patch 1
- install patch 2
- fail boot patch 2
- fail boot patch 1 (ignore failure)
- keep trying patch 1 since we succeeded once.

I'm fine with us having a "we will roll back until we successfully boot" behavior,

What about?
- success boot patch 1
- fail boot patch 1
Do we roll back in that case?  Previously I was under the impression we should not.

Regardless, I'd encourage testing all of the above.

bryanoltman: The first case you mention isn't exactly right, as we're checking here whether we should even try to boot patch 1 as a fallback. What we currently try to handle now is:

Case A:
1. Patch 1 boot success
2. Patch 2 boot failure
3. Patch 1 still bootable?
4. If so, try boot patch 1
5. If not, fall back to release

> Do we roll back in that case? Previously I was under the impression we should not.

Yes, we do and should. Otherwise we'd end up repeatedly trying to boot patch 1 and effectively brick the app.

Calling this scenario "Case B":
- success boot patch 1
- fail boot patch 1

We test both cases between the `fall_back_tests` and the `next_boot_patch_tests` (because `next_boot_patch` is the public API that exercises this code):

- Case A:
  - `falls_back_to_last_booted_patch_if_still_bootable`
  - `does_not_fall_back_to_last_booted_patch_if_corrupted`
  - `clears_next_and_last_patches_if_both_fail_validation`
- Case B:
  - `returns_none_if_next_boot_patch_is_not_bootable`, 
  - `clears_current_and_next_on_boot_failure_if_they_are_the_same` 

Let me know if you feel like something is missing here. AFAICT, we test all fallback behavior cases and verify that the correct artifacts are left on disk afterwards in the tests as they currently stand.

eseidel: Thank you for the thorough response.  This is definitely fodder for some later writings on safety & security.

eseidel (approved):

PR #150, opened 2024-04-23
fix: delete all old patch artifacts on successful boot

bryanoltman:
## Description

On successful boot, deletes all patch artifacts belonging to patches older than the successfully booted patch.

Also changes `record_boot_success_for_patch` to `record_boot_success`, as the patch manager now knows which patch we're recording success for.

Fixes https://github.com/shorebirdtech/updater/issues/147

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -837,6 +837,7 @@ mod fall_back_tests {
         manager.add_patch_for_test(&temp_dir, 1)?;
 
         // Start booting from patch 1.
+        manager.record_boot_start_for_patch(1)?;
 
         // Download patch 2 before patch 1 finishes booting.
         manager.add_patch_for_test(&temp_dir, 2)?;
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -195,25 +195,30 @@ impl PatchManager {
             .with_context(|| format!("Failed to delete patch dir {}", &patch_dir.display()))
     }
 
-    /// Attempts to use the last successfully booted patch as the next boot patch. If the last successfully
-    /// booted patch is not bootable or has the same number as the patch we're falling back from, we clear it.
-    fn try_fall_back_to_last_booted_patch(&mut self) {
-        if let Some(next_boot_patch) = self.patches_state.next_boot_patch {
-            // If we have a next_boot_patch that we're falling back from, delete its artifacts.
-            let _ = self.delete_patch_artifacts(next_boot_patch.number);
-            self.patches_state.next_boot_patch = None;
+    /// Deletes artifacts for the provided bad_patch_number and attempts to set the next_boot_patch to the last
+    /// successfully booted patch. If the last successfully booted patch is not bootable or has the same number
+    /// as the patch we're falling back from, we clear it as well.
+    fn try_fall_back_from_patch(&mut self, bad_patch_number: usize) {
+        let _ = self.delete_patch_artifacts(bad_patch_number);
 
-            if let Some(last_boot_patch) = self.patches_state.last_booted_patch {
-                if last_boot_patch.number == next_boot_patch.number {
-                    // If the last booted patch is the same as the next boot patch, clear it.
-                    self.patches_state.last_booted_patch = None;
+        if let Some(next_boot_patch) = self.patches_state.next_boot_patch {
+            // If our next boot patch is bad_patch_number, clear it.
+            if next_boot_patch.number == bad_patch_number {
+                self.patches_state.next_boot_patch = None;
+
+                // If the last booted patch is the same as the next boot patch, clear it.
+                if let Some(last_boot_patch) = self.patches_state.last_booted_patch {
+                    if last_boot_patch.number == bad_patch_number {
+                        self.patches_state.last_booted_patch = None;
+                    }
                 }
             }
         }
 
+        // If we think we can still boot from the last booted patch, set it as the next_boot_patch.
+        // If something happened to render the last boot patch unbootable, clear it and delete its artifacts.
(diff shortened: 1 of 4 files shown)


library/src/cache/patch_manager.rs
```diff
@@ -70,7 +70,7 @@ pub trait ManagePatches {
 
     /// Records that the patch with number patch_number booted successfully and is
     /// safe to use for future boots.
-    fn record_boot_success_for_patch(&mut self, patch_number: usize) -> Result<()>;
+    fn record_boot_success(&mut self) -> Result<()>;
```
eseidel, line 73: That means this is moving from being a "stateless" to "stateful" api, which may be fine.  But probably as part of that you'd want the patch_manager to always be created with a specific current_patch?  And then never allow that to be changed?

Otherwise you might have the problem of not easily being able to intuit from the call sites as to what "current_patch" they're affecting?

library/src/cache/patch_manager.rs
```diff
@@ -275,6 +275,19 @@ impl PatchManager {
             }
         }
     }
+
+    /// Deletes all patch artifacts with numbers less than patch_number.
+    fn delete_patch_artifacts_older_than(&mut self, patch_number: usize) -> Result<()> {
```
eseidel, line 280: We've talked about moving away from patch_number (which I think it just a name?) to some sort of patch_id?  Obviously that's not the state today.  Obviously we will still need some way to compare them by order.

library/src/cache/patch_manager.rs
```diff
@@ -275,6 +275,19 @@ impl PatchManager {
...
+        for entry in std::fs::read_dir(self.patches_dir())? {
+            let entry = entry?;
+            let number = entry.file_name().to_string_lossy().parse::<usize>()?;
+            if number < patch_number {
+                let _ = self.delete_patch_artifacts(number);
+            }
```
eseidel, line 286: Do you want this API to just delete all others? Is there ever a case where you want to keep around "newer" ones?

library/src/cache/patch_manager.rs
```diff
@@ -367,28 +380,14 @@ impl ManagePatches for PatchManager {
...
-            );
-        }
-
-        if let Some(current_patch) = self.patches_state.last_booted_patch {
-            if current_patch.number != patch_number {
-                // If we now have a new last_booted_patch, delete the old one's artifacts.
-                let _ = self.delete_patch_artifacts(current_patch.number);
-            }
-        }
+            .last_attempted_patch
```
eseidel, line 386: Is "last_attempted_patch" the "current patch"?

library/src/cache/patch_manager.rs
bryanoltman: I'm not sure I recall – was this to allow iOS and Android to each have sequential patch numbers?

library/src/cache/patch_manager.rs
bryanoltman: We want to keep around newer patches. Consider:

1. We start booting patch 2
2. While booting (i.e., in between boot start and boot success), we download and inflate patch 3
3. We finish booting patch 2

Deleting all other patch artifacts would delete patch 3, and because we've "seen" patch 3, we would never try to download it again (it would be considered "bad").

Adding a comment.

library/src/cache/patch_manager.rs
bryanoltman: In this state, yes. This is the implementation of the `ManagePatches` trait, so all of the docs are above.

library/src/cache/patch_manager.rs
eseidel: To allow them to have the same patch number I believe.

library/src/cache/patch_manager.rs
```diff
@@ -277,6 +277,27 @@ impl PatchManager {
...
+    /// 1. We start booting patch 2
+    /// 2. While booting (i.e., in between boot start and boot success), we download and inflate patch 3
+    /// 3. We finish booting patch 2
+    ///
+    /// Deleting all other patch artifacts would delete patch 3, and because we've "seen" patch 3,
+    /// we would never try to download it again (it would be considered "bad").
+    fn delete_patch_artifacts_older_than(&mut self, patch_number: usize) -> Result<()> {
+        for entry in std::fs::read_dir(self.patches_dir())? {
+            let entry = entry?;
+            let number = entry.file_name().to_string_lossy().parse::<usize>()?;
```
eseidel, line 293: What happens if they ended up with junk in this directory?

library/src/cache/patch_manager.rs
```diff
@@ -369,28 +390,14 @@ impl ManagePatches for PatchManager {
...
-                // If we now have a new last_booted_patch, delete the old one's artifacts.
-                let _ = self.delete_patch_artifacts(current_patch.number);
-            }
-        }
+            .last_attempted_patch
+            .context("No last_attempted_patch")?;
 
-        self.patches_state.last_booted_patch = Some(next_boot_patch);
+        self.patches_state.last_booted_patch = Some(boot_patch);
+        let _ = self.delete_patch_artifacts_older_than(boot_patch.number);
```
eseidel, line 400: perhaps document why we're ignoring the error?

eseidel (approved):

library/src/cache/patch_manager.rs
bryanoltman: Yep, that's what I meant. I think the shift from "patch number as identifier" to some other ordered patch identifier will be pretty straightforward (maybe just `platformN` for patch number `N` on `platform`?) but definitely a bigger project than what I'm doing here.

eseidel (approved):

PR #151, opened 2024-04-24
docs: update state machine description in readme

bryanoltman:
## Description

Updates the README to reflect the current state of the Updater state machines.

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/README.md
+++ b/library/README.md
@@ -124,9 +124,9 @@ NDK_HOME=$HOME/Documents/GitHub/engine/src/third_party/android_tools/ndk cargo n
 #### Boot State Machine
 
 This state machine tracks the process of the engine starting up, with or without
-a patch. These state transitions happen with or without a patch, but in the
-context of the updater, we only care about the case where we are booting from
-a patch.
+a patch. These state transitions happen whether or not a new patch is available,
+but in the context of the updater, we only care about the case where we are
+booting from a patch.
 
 The patch boot state is internal to the PatchManager and stored on disk in
 `patches_state.json`. It contains three fields, all of which are Optional
--- a/library/README.md
+++ b/library/README.md
@@ -119,29 +119,84 @@ NDK_HOME=$HOME/Documents/GitHub/engine/src/third_party/android_tools/ndk cargo n
 - Updates must be applied in order.
 - Updates are applied in a single transaction.
 
-### Update State Machine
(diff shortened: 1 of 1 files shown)


library/README.md
```diff
@@ -119,35 +125,90 @@ NDK_HOME=$HOME/Documents/GitHub/engine/src/third_party/android_tools/ndk cargo n
...
+- last_attempted_patch: The last patch that we attempted to boot.
+- next_boot_patch: The next patch that we will attempt to boot.
+
+This state machine has the following states. It can only move forward through
+them.
+
+1. Ready - The engine is initialized but has not started booting yet.
+2. Booting - The engine has started booting.
+3. Success - The engine successfully booted.
+4. Failure - The engine failed to boot.
```
eseidel, line 151: Is this actually recorded in an enum?  (if not that might be a next step, or worth commenting).

library/README.md
```diff
@@ -119,35 +125,90 @@ NDK_HOME=$HOME/Documents/GitHub/engine/src/third_party/android_tools/ndk cargo n
...
+- `record_launch_failure`: Moves from Booting to Failure
+  - next_boot_patch artifacts are deleted.
+  - next_boot_patch is set to either:
+    - last_boot_patch if it is still valid, or
+    - None, if last_boot_patch is None or invalid.
+
+These are effectively no-ops if we are not booting from a patch.
+
+We assume that this state machine will have advanced at least as far as the
+Booting state before the Patch Check State Machine (below) is started.
```
eseidel, line 171: Is any call to an earlier method an error?  Should it be?

eseidel (approved):
This is fantastic!  I think the code should just do this now and would be more clear. :)

library/README.md
bryanoltman: Added an assumptions list at the bottom and specifically called out that PatchManager state is implicit, but should be more explicit.

library/README.md
bryanoltman: This also became part of the "not enforced but should be" assumptions list above.

bryanoltman:
> I think the code should just do this now and would be more clear

It does :) I think the fact that _that_ isn't clear should probably be addressed though 🙃

PR #155, opened 2024-05-03
reading public key from release artifact

erickzanardo:
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -1,3 +1,5 @@
+use crate::config::with_config;
+
 use super::{disk_io, PatchInfo};
 use anyhow::{bail, Context, Result};
 use base64::Engine;
@@ -184,6 +186,7 @@ impl PatchManager {
     ///
     /// Returns Ok if the patch is bootable, or an error if it is not.
     fn validate_patch_is_bootable(&self, patch: &PatchMetadata) -> Result<()> {
+
         let artifact_path = self.patch_artifact_path(patch.number);
         if !Path::exists(&artifact_path) {
             bail!(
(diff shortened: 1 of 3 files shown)


library/src/yaml.rs
```diff
@@ -12,6 +12,8 @@ pub struct YamlConfig {
     pub base_url: Option<String>,
     /// Update behavior. Defaults to true if not set.
     pub auto_update: Option<bool>,
+    /// Path for the public key to verify the patch.
```
bryanoltman, line 15:
```suggestion
    /// Path to a public key that will be used to verify patch hash signatures.
```

bryanoltman: Also, what format are we expecting this to be in?

library/src/config.rs
```diff
@@ -128,6 +129,7 @@ pub fn set_config(
                 .to_owned(),
             network_hooks,
             file_provider,
+            patch_public_key_path: yaml.patch_public_key_path.clone(),
```
bryanoltman, line 132: This should probably use `toString()` for consistency with the above.

library/src/cache/patch_manager.rs
```diff
@@ -225,54 +228,67 @@ impl PatchManager {
             }
         }
 
-        // Ensure patch signature is valid for recorded hash
+        let result = with_config(|config| {
```
bryanoltman, line 231: We should pass in the public key instead of using this

library/src/yaml.rs
erickzanardo: This is just a string path of the public key file for now. Did you had something else in mind?

bryanoltman: Right, what should that file contain? Raw bytes? A base64-encoded string?

erickzanardo: Ah yes, for my testing I am using that .pem file. Which I believe is just a text file?

I am not sure though what would be the best case though?

bryanoltman: Yep, just document it in the code :)

library/src/yaml.rs
```diff
@@ -12,6 +12,8 @@ pub struct YamlConfig {
     pub base_url: Option<String>,
     /// Update behavior. Defaults to true if not set.
     pub auto_update: Option<bool>,
+    /// Path to a public key that will be used to verify patch hash signatures.
+    pub patch_public_key_path: Option<String>,
```
eseidel, line 16: I'm confused by this path because shorebird.yaml on disk vs. shorebird.yaml in the apk are different formats.

Does this point to the path on disk during development and an apk path during production?

library/src/cache/patch_manager.rs
```diff
@@ -184,6 +186,7 @@ impl PatchManager {
     ///
     /// Returns Ok if the patch is bootable, or an error if it is not.
     fn validate_patch_is_bootable(&self, patch: &PatchMetadata) -> Result<()> {
+
```
eseidel, line 189:
```suggestion
```

library/src/cache/patch_manager.rs
```diff
@@ -225,54 +227,67 @@ impl PatchManager {
...
 
-        Ok(())
+          Ok(())
+        });
+
+        // Idk why this is needed
+        match result {
+            Ok(_) => { Ok(()) }
+            Err(e) => bail!(e)
+        }
```
eseidel, line 290: It isn't.  You can just use `?` on the line where `result` is computed.

erickzanardo:
@bryanoltman given all the chats and discussion we have been having on the path that the code sign is heading, I feel we are safe to close this POC right? I don't think there is much we could salvage from here

bryanoltman:
I think that's fair

PR #157, opened 2024-05-08
docs: add thoughts on future design

eseidel:


README.md
```diff
@@ -16,11 +16,16 @@ into the Flutter engine during build time.
...
+- `library`: Runtime library linked into Flutter Engine to unpack and apply
+  updates.  Build into libupdater.a and linked into libflutter.so.
+- `patch`: Developer tooling to package Shorebird updates ("patches"). Built
+  into `patch.exe` and downloaded and run by `shorebird` command line tool:
+  https://github.com/shorebirdtech/shorebird/tree/main/packages/shorebird_cli.
+- `shorebird_code_push`: The Dart bindings for communicating with the updater
+  library from within a Flutter app.  Published to pub.dev, usage is optional by
+  developers.
+
+Mos interesting code is in the `library` directory. There is also
```
bryanoltman, line 28:
```suggestion
Most interesting code is in the `library` directory. There is also
```

erickzanardo, line 28:
```suggestion
Most interesting code is in the `library` directory. There is also
```

patch/README.md
```diff
@@ -1,21 +1,40 @@
 # patch command line tool
 
-This is the tool used by the `shorebird` command line to compute the patch
-file for uploading to the server.
-
-This currently uses the rust `bidiff` crate to compute the patch file.
-and could just use the `bic` command line tool included in that crate. However
-we're explicitly writing our own command line to allow us to change the
-underlying compression without affecting the `shorebird` command line callers.
+This is the tool used by the `shorebird` command line to package up a patch
```
bryanoltman, line 3: We might even just say this creates a binary diff (I believe that's all it does?)

patch/README.md
```diff
@@ -1,21 +1,40 @@
...
+`bidiff` dependencies eventually.
+
+Because `bidiff` does not check what it's patching, it's possible to apply a
+patch to the wrong file. To avoid this we currently have a separate hash
+which we save in Shorebird's database and then `library` validates that the
+patched file matches what we expected after inflation.  This is the wrong design
+and we intend to move to a system whereby `patch` is responsible for including
+its own hash in the patch file and validating during application.
+
+In that world we might end up with two hashes.  One who's purpose is to validate
```
bryanoltman, line 26:
```suggestion
In that world we might end up with two hashes.  One whose purpose is to validate
```

patch/README.md
bryanoltman: nvm, answered later

patch/README.md
erickzanardo, line 26:
```suggestion
In that world we might end up with two hashes. One who's purpose is to validate
```

PR #161, opened 2024-05-13
fix: Don't make network request while holding config lock

bryanoltman:
Update `check_for_update_internal` to no longer hold the config lock while making a patch check request.

Fixes https://github.com/shorebirdtech/shorebird/issues/1981

### DIFF
--- a/library/src/updater.rs
+++ b/library/src/updater.rs
@@ -533,7 +533,7 @@ mod tests {
 
     use crate::{
         config::{testing_reset_config, with_config},
-        network::{testing_set_network_hooks, NetworkHooks},
+        network::{testing_set_network_hooks, NetworkHooks, PatchCheckResponse},
         ExternalFileProvider,
     };
 
@@ -835,8 +835,11 @@ mod tests {
         .unwrap();
     }
 
+    #[serial]
     #[test]
     fn no_config_lock_contention_when_waiting_for_patch_check() {
+        static mut HAS_FINISHED_CONFIG: bool = false;
+
         let tmp_dir = TempDir::new("example").unwrap();
         init_for_testing(&tmp_dir, None);
 
@@ -845,7 +848,16 @@ mod tests {
             patch_check_request_fn: |_url, _request| {
                 let patch_check_delay = std::time::Duration::from_secs(2);
                 std::thread::sleep(patch_check_delay);
-                // If we get here, the test has failed.
+
+                // If we've obtained and released the config lock, this test has passed.
+                if unsafe { HAS_FINISHED_CONFIG } {
+                    return Ok(PatchCheckResponse {
+                        patch_available: false,
+                        patch: None,
+                    });
+                }
+
+                // If we have not yet finished with the config lock, this test has failed.
                 unreachable!("If the test has not terminated before this, set_config is likely being blocked by a patch check request, which should not happen");
             },
             download_file_fn: |_url| Ok([].to_vec()),
@@ -859,16 +871,14 @@ mod tests {
         );
 
         // Invoke check_for_update to kick off a patch check request
-        let _ = std::thread::spawn(|| {
-            crate::check_for_update().unwrap();
-        });
+        let _ = std::thread::spawn(crate::check_for_update);
 
         // Call with_config to get the config lock. This should complete before the patch check request is resolved.
-        let config_thread = std::thread::spawn(|| {
-            let _ = with_config(|_| Ok(()));
-        });
+        let config_thread = std::thread::spawn(|| with_config(|_| Ok(())));
 
-        config_thread.join().unwrap();
+        assert!(config_thread.join().is_ok());
+
+        unsafe { HAS_FINISHED_CONFIG = true };
 
         // Don't wait for the patch check thread. This test should finish more or less immediately
         // if with_config isn't waiting for the patch check thread. If it is waiting, this test will
--- a/library/src/updater.rs
+++ b/library/src/updater.rs
@@ -506,7 +506,11 @@ mod tests {
     use std::fs;
     use tempdir::TempDir;
 
-    use crate::{config::testing_reset_config, ExternalFileProvider};
+    use crate::{
+        config::{fake_yaml, set_config, testing_reset_config},
+        network::{testing_set_network_hooks, NetworkHooks, PatchCheckResponse},
+        ExternalFileProvider,
+    };
 
     #[derive(Debug, Clone)]
     struct FakeExternalFileProvider {}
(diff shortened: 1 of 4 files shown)


library/src/updater.rs
```diff
@@ -144,14 +145,38 @@ pub fn should_auto_update() -> anyhow::Result<bool> {
...
+        Ok((
+            patch_check_request(config, &state),
+            patches_check_url(&config.base_url),
+            config.network_hooks.patch_check_request_fn,
+        ))
+    })?;
+
+    let response = request_fn(&url, request)?;
+    debug!("Patch check response: {:?}", response);
+    Ok(response)
```
eseidel, line 179: How does the response get recorded?  I guess the caller does that?

library/src/updater.rs
```diff
@@ -805,4 +834,68 @@ mod tests {
...
+                );
+            }
+        });
+
+        set_config_thread.join().unwrap();
+
+        // Don't wait for the patch check thread. This test should finish more or less immediately
+        // if set_config isn't waiting for the patch check thread. If set_config is waiting for
+        // the patch check thread, this test will not finish for the duration of patch_check_delay
+        // (defined in patch_check_request_fn above).
```
eseidel, line 899: Could probably use a global variable to ensure that we get to the end of the test before the network callback is triggered?

library/src/updater.rs
eseidel: I guess the response never is recorded? TIL.

bryanoltman: Nope, we don't log the response. What's interesting is that this function is not called by the engine – seemingly only ever by the `shorebird_code_push` package.

eseidel: I think that's why check_for_update_internal exists, was to share code between this (new) function which was exposed for package:shorebird_code_push, and the normal update mechanism.

As I look back through this code I am reminded of how simple/primitive it is.  We just haven't touched it much in a year.

That also might mean that what's happening is that package:shorebird_code_push is running already in a background thread while the user is launching the app (shorebird_init?)

I suspect that we'll also (possibly separately) want to do the fix where we make set_config be able to return quickly without even trying to grab the config lock to help shorebird_init always be fast?  Or maybe we just need to be extra careful to make sure we're never holding the config lock for any length of time (e.g. across a network request).

eseidel:
The bug you're fixing is basically making it so that in some circumstances it's possible to cause application launch to hang waiting on a network request (which is a terrible thing for any app to do).  Obviously unintentional on our part.

library/src/updater.rs
bryanoltman: Did you one better – there's an `unreachable` in the patch check callback after we sleep, which will cause the test to fail if we're waiting for a response and shouldn't be.

library/src/updater.rs
bryanoltman: > We just haven't touched it much in a year.

Yep, it's been relatively stable + the consequences of getting it wrong are quite high.

> I suspect that we'll also (possibly separately) want to do the fix where we make set_config be able to return quickly without even trying to grab the config lock to help shorebird_init always be fast

I think we could spawn a new thread when calling `set_config` from `init` to make it look something like:

```rust
pub fn init(
    app_config: AppConfig,
    file_provider: Box<dyn ExternalFileProvider>,
    yaml: &str,
) -> Result<(), UpdateError> {
    #[cfg(any(target_os = "android", test))]
    use crate::android::libapp_path_from_settings;

    init_logging();
    let config = YamlConfig::from_yaml(yaml)
        .map_err(|err| UpdateError::InvalidArgument("yaml".to_string(), err.to_string()))?;

    let libapp_path = libapp_path_from_settings(&app_config.original_libapp_paths)?;
    debug!("libapp_path: {:?}", libapp_path);

    let _ = std::thread::spawn(move || {
        set_config(
            app_config,
            file_provider,
            libapp_path,
            &config,
            NetworkHooks::default(),
        );
    });

    Ok(())
}
```

patch/src/main.rs
```diff
@@ -23,7 +23,7 @@ fn main() {
         eprintln!("  base:   Path to the base file");
         eprintln!("  new:    Path to the new file");
         eprintln!("  output: Path to the output patch file");
-        eprintln!("");
```
bryanoltman, line 26: This is just to get the linter to stop yelling

library/src/updater.rs
eseidel: I don't think you want that.  Any time the lock is already taken, there is nothing for us to do.

What about using https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.try_lock in the set_config case?  If the lock is already taken, we know that config is initialized and thus nothing to do (since any other taking of the lock when we're not initialized errors I believe).

bryanoltman: That would work. We can discuss in the future PR :)

library/src/updater.rs
```diff
@@ -805,4 +834,45 @@ mod tests {
...
+        let tmp_dir = TempDir::new("example").unwrap();
+        init_for_testing(&tmp_dir, None);
+
+        // Set up the network hooks to sleep for 10 seconds on a patch check request
+        let hooks = NetworkHooks {
+            patch_check_request_fn: |_url, _request| {
+                let patch_check_delay = std::time::Duration::from_secs(2);
+                std::thread::sleep(patch_check_delay);
+                // If we get here, the test has failed.
+                unreachable!("If the test has not terminated before this, set_config is likely being blocked by a patch check request, which should not happen");
```
eseidel, line 849: Does the test terminating cause the check_for_update thread to get killed?  I'm not sure what code here is tearing down the thread before this gets hit?  A naive implementation I would assume this to just get hit *after* the test runs (while some other test is running).

library/src/updater.rs
```diff
@@ -805,4 +834,45 @@ mod tests {
...
+
+        testing_set_network_hooks(
+            hooks.patch_check_request_fn,
+            hooks.download_file_fn,
+            hooks.report_event_fn,
+        );
+
+        // Invoke check_for_update to kick off a patch check request
+        let _ = std::thread::spawn(|| {
+            crate::check_for_update().unwrap();
```
eseidel, line 863: I think you want .ok() instead of .unwrap().  Unwrap will just ignore the error, right?  Where as ok() will panick if it errored?

eseidel (changes requested):
I think this needs one more round (to explain why the unreachable is never hit, confirm that it's actually doing something) and confirm .unwrap() isn't just silently hiding failures.

library/src/updater.rs
bryanoltman: > Does the test terminating cause the check_for_update thread to get killed?

No, although I'm not sure what happens if the tests take longer than 2s to run. Let me dig into that a bit more.

library/src/updater.rs
eseidel: Apologies for swapping the two. Thanks!

library/src/updater.rs
bryanoltman: Added a bool to ensure that we don't hit `unreachable` if we're managed to obtain and release the lock

bryanoltman:
I think I've addressed all feedback, although maybe not

> confirm that it's actually doing something

What do you mean by that?

eseidel:
> I think I've addressed all feedback, although maybe not
> 
> > confirm that it's actually doing something
> 
> What do you mean by that?

Because I was unclear as to how the thread actually gets canceled (presumably rust has some fancy built in way?) it wasn't clear to me that the network callback was ever called.  If I were writing this I might just remove the 2s timeout and see if the test then correctly crashes to your unreachable. 🤷   If you tell me it's working I believe you.

eseidel (approved):
lgtm.

Please confirm, the test failed before you change, correct?

bryanoltman:
> If I were writing this I might just remove the 2s timeout and see if the test then correctly crashes to your unreachable.

I could probably shorten the delay, but removing it makes the test flaky—the config thread and the patch check thread are spawned close enough together that it seems to be a coin flip whether the `patch_check_request_fn` body executes before `HAS_FINISHED_CONFIG = true` is set.

bryanoltman:
> Please confirm, the test failed before you change, correct?

Confirmed. If I swap out the body of `check_for_update_internal` with the following, this test fails

```rust
    with_config(|config| {
        // Load UpdaterState from disk
        // If there is no state, make an empty state.
        let state =
            UpdaterState::load_or_new_on_error(&config.storage_dir, &config.release_version);
        send_patch_check_request(config, &state)
    })
```

eseidel:
It's always possible to de-flake threads with mutexs of course if you want to go that route.

Most important is that the test fails (even flakily) before your change and passes consistently after.

PR #163, opened 2024-05-16
feat: add updater_tools with package_patch command

bryanoltman:
## Description

Adds an initial `updater_tools package_patch` command that accepts two aabs (a release and a patch) and generates patch artifacts for them.

Sample usage:

```
dart run updater_tools package_patch \
  --archive-type=aab \
  --release ~/Desktop/release.aab \
  --patch ~/Desktop/patch.aab \
  --patch-executable=$HOME/shorebirdtech/_shorebird/shorebird/bin/cache/artifacts/patch/patch \
  --output=$HOME/Desktop/testout 
```

Outputs to `$HOME/Desktop/testout`:

```
bryanoltman@moltman ~/Desktop
⑆ l testout
total 24
drwx------   5 bryanoltman  staff   160B May 16 11:55 .
drwx------@ 88 bryanoltman  staff   2.8K May 16 11:55 ..
-rw-r--r--   1 bryanoltman  staff   160B May 16 11:54 arm64-v8a.zip
-rw-r--r--   1 bryanoltman  staff   160B May 16 11:54 armeabi-v7a.zip
-rw-r--r--   1 bryanoltman  staff   160B May 16 11:55 x86_64.zip
```

Where the contents of each .zip file is the `dlc.vmcode` produced by the `patch` executable.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


updater_tools/lib/src/commands/package_patch_command.dart
```diff
@@ -0,0 +1,145 @@
...
+import 'package:updater_tools/src/artifact_type.dart';
+import 'package:updater_tools/src/commands/updater_tool_command.dart';
+import 'package:updater_tools/src/logger.dart';
+import 'package:updater_tools/src/packager/patch_packager.dart';
+
+/// The arg name to specify the release and patch archive type.
+const archiveTypeCliArg = 'archive-type';
+
+/// The arg name to specify the path to the release archive.
+const releaseCliArg = 'release';
```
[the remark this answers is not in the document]
bryanoltman: Why yes I did! Fixing

updater_tools/lib/src/commands/package_patch_command.dart
```diff
@@ -0,0 +1,145 @@
...
+    final patchFilePath = results['patch'] as String;
+    final patchExecutablePath = results['patch-executable'] as String;
+
+    if (!File(releaseFilePath).existsSync()) {
+      throw ArgumentError.value(
+        releaseFilePath,
+        'release',
+        'The release file does not exist',
+      );
+    }
```
eseidel, line 127: I wonder if you might want to just make these a helper function?

verifyFilePath(argResults, 'release');

bryanoltman: Good call, done

updater_tools/lib/src/packager/patch_packager.dart
```diff
@@ -0,0 +1,175 @@
...
+
+  /// Create and package a patch of [patchArchive] onto [releaseArchive].
+  Future<Directory> packagePatch({
+    required File releaseArchive,
+    required File patchArchive,
+    required ArchiveType archiveType,
+    required Directory outputDirectory,
+  }) async {
+    final directory = switch (archiveType) {
+      ArchiveType.aab => await _packageAndroidAabPatch(
```
felangel, line 50: nice 💯

felangel (approved):
Overall looks great. One thing to consider is the format of the contents of each .zip as well as the contents of the output directory may change over time and it might be nice to have some sort of manifest which makes this explicit otherwise each time we change the output format or add new files, we'll need to support checking updater_tools versions etc. to decide how to interpret the output contents.

bryanoltman:
The contents of each zip is intended to be opaque, the idea being that `arch_name.zip` will be what is handed back to the CLI and (eventually) to the updater as a patch artifact. In theory, no other part of the system should need to care about the contents/layout of the zip.

Re: the content/layout of the output directory, I think that should be stable so long as we stick to a one-patch-artifact-per-arch model. We can always ask `updater_tools` what its version is and skew on that if/when we need as well.

PR #171, opened 2024-05-21
feat: update patch manager to check signature on boot

bryanoltman:
## Description

Update PatchManager to, if a public key is present, verify the signature of an inflated patch's hash before booting the patch.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


erickzanardo (approved):

library/src/cache/patch_manager.rs
```diff
@@ -229,6 +234,20 @@ impl PatchManager {
...
+                .signature
+                .clone()
+                .context("Patch signature is missing")?;
+
+            // Check that the signature is valid.
+            let patch_hash = signing::hash_file(&artifact_path)?;
+            signing::check_signature(&patch_hash, &patch_signature, public_key)?;
+        } else {
+            info!("No public key provided, skipping signature verification");
+        }
```
eseidel, line 249: Not sure where the hash check is.

I would expect the process to be:
1. If we have a public key, check the signature on the hash matches.
2. Hash the patch
3. Check the hash matches.

Presumably that's already being done, just spread across other methods?

library/src/cache/patch_manager.rs
```diff
@@ -229,6 +234,20 @@ impl PatchManager {
             }
         }
 
+        if let Some(public_key) = &self.patch_public_key {
+            // If we have a public key, verify that the patch has a signature
+            let patch_signature = patch
```
eseidel, line 239: It's really the hash that's signed rather than the whole patch, correct?

library/src/cache/patch_manager.rs
```diff
@@ -810,6 +845,95 @@ mod next_boot_patch_tests {
...
+
+        manager.add_signed_patch_for_test(&temp_dir, 1, INFLATED_PATCH_HASH, Some(SIGNATURE))?;
+
+        assert!(manager.next_boot_patch().is_none());
+
+        Ok(())
+    }
+
+    #[test]
+    fn returns_none_if_patch_is_missing_expected_signature() -> Result<()> {
```
eseidel, line 875: This is also "fails" right?  Just a different flavor of return?  Both this and the above tests refuse to boot, correct?

library/src/cache/patch_manager.rs
```diff
@@ -810,6 +845,95 @@ mod next_boot_patch_tests {
...
+
+        assert!(manager.next_boot_patch().is_some());
+        let patch = manager.next_boot_patch().unwrap();
+        assert_eq!(patch.number, 1);
+
+        Ok(())
+    }
+
+    #[test]
+    fn returns_patch_with_arbitrary_signature_if_no_public_key() -> Result<()> {
```
eseidel, line 919: Tests I can imagine:
1. no public key + signed patch = boots
1. no public key + unsigned patch = boots
1. no public key + invalid signature = ?
1. public key + signed patch = boots
1. public key + unsigned patch = fails
1. public key + invalid signature = fails

library/src/cache/patch_manager.rs
bryanoltman: > Not sure where the hash check is.

We don't currently check the hash for equality, and I'm not sure we need to—any change to the hash would cause the signature to become invalid.

The process is:

1. If we have a public key, hash the patch file
2. Check that the signature matches the hash

library/src/cache/patch_manager.rs
bryanoltman: Yep. The contract is that `next_boot_patch` will only return a patch if there is a patch that we think is valid. If there's a patch but we think it's invalid for any reason, `next_boot_patch` returns None.

library/src/cache/patch_manager.rs
bryanoltman: > no public key + invalid signature = ?

I'm not sure what it would mean for a signature to be invalid with no public key.

Going through the other items on the list:

- [x] no public key + signed patch = boots (`returns_patch_with_arbitrary_signature_if_no_public_key`)
- [x] no public key + unsigned patch = boots (all previously existing tests)
- [x] public key + signed patch = boots (`returns_patch_if_patch_has_valid_signature`)
- [x] public key + unsigned patch = fails (`returns_none_if_patch_is_missing_expected_signature`)
- [x] public key + invalid signature = fails (`returns_none_if_patch_has_invalid_signature`)

library/src/cache/patch_manager.rs
bryanoltman: Yep, updated to reflect this

library/src/cache/patch_manager.rs
eseidel: > I'm not sure what it would mean for a signature to be invalid with no public key.

It could go either way.  Either any signature failure (e.g parsing the signature) causes patch failure, or the signature is completely ignored when there is no public key.  Both would be valid behaviors.

library/src/cache/patch_manager.rs
eseidel: I think I/we need to write out exactly what we're singing in plain language.  Maybe I'll add that to my in-progress blog post tomorrow.

library/src/cache/patch_manager.rs
```diff
@@ -810,6 +845,95 @@ mod next_boot_patch_tests {
 
         Ok(())
     }
+
+    // The constant values below were generated by taking the hash of the patch file
+    // using openssl to sign it with a private key.
```
eseidel, line 850: What is "the patch file"? here?  "MESSAGE"?

library/src/cache/signing.rs
```diff
@@ -0,0 +1,133 @@
...
+        .with_context(|| format!("Failed to decode public_key: {}", public_key))?;
+
+    let public_key =
+        signature::UnparsedPublicKey::new(&signature::RSA_PKCS1_2048_8192_SHA256, public_key_bytes);
+    let decoded_sig = match base64::prelude::BASE64_STANDARD.decode(signature) {
+        Ok(sig) => sig,
+        Err(e) => {
+            bail!("Failed to decode signature: {:?}", e);
+        }
+    };
```
eseidel, line 44: Is this just .context()?

eseidel: Or .map_err()?

bryanoltman: Possibly could be, let me see if there's a way to include the error message in a `.context`

eseidel (approved):

erickzanardo (approved):

library/src/cache/patch_manager.rs
bryanoltman: Improved these a bit

library/src/cache/signing.rs
bryanoltman: `map_err` works, just had to use `anyhow`'s Error type, which is what is "thrown" by `bail!`:

```rust
    let decoded_sig = base64::prelude::BASE64_STANDARD
        .decode(signature)
        .map_err(|e| anyhow::Error::msg(format!("Failed to decode signature: {:?}", e)))?;
```

Issue #184, opened 2024-07-13
Accessing next_boot_patch() during patch launch marks patch as invalid


bryanoltman:
`PatchManager`'s `next_boot_patch` function validates the patch in the "next boot" slot before returning it to avoid attempting to boot an invalid patch. If a patch is invalid, it will not return that patch (either a previous patch or `None` will be returned instead). One way a patch can be found invalid is if we have a record of attempting to boot from it but no record of successfully booting from it.

This method of validation leaves open a window where a new patch that has started booting but has not yet recorded a successful boot. Calls to `next_boot_patch` during this window will run validation on the booting patch (because it is in the `next_boot_patch` slot) and it will fail validation, causing it to be deleted in the boot process.

eseidel:
Eek!  Presumably it’s possible for customers to hit this via the
shorebird_code_push package?

bryanoltman:
I stumbled onto this because of a change in https://github.com/shorebirdtech/updater/pull/183 (using `next_boot_patch` instead of `latest_seen_patch_number` when constructing a patch check request). It looks like calling `shorebirdCodePush.nextPatchNumber()` as the first line of a Flutter app's `main` does not trigger this (tested on an Android emulator), so I imagine the likelihood of someone hitting this because of `shorebird_code_push` is extremely low if not nil.

eseidel:
Ah. Yeah our current “record successful” runs right after dart successfully
initializes. We should probably run it after we reach main or the first
frame and if we had done that we would have similarly hit this. Thanks.

PR #185, opened 2024-07-15
fix: track patches while they are booting

bryanoltman:
## Description

When validating that a patch is "good" to boot from, we compare `last_attempted_boot_patch` to `last_successfully_booted_patch`. If these are different, we assume that `last_attempted_boot_patch` failed to boot and delete it. This causes problems if we perform validation while that patch is booting (we are attempting to boot from it, but it has not yet finished booting, so we think it is bad and delete it, never giving it a chance to finish booting).

To get around this, this PR introduces `currently_booting_patch`, which causes us to skip the above step of validation and prevents us from assuming that a patch in the process of booting has previously failed to boot.

As a followup to this, we could probably delete `last_attempted_boot_patch` and use `on_init` to clean up a "failed to boot" patch.

Fixes https://github.com/shorebirdtech/updater/issues/184

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -899,7 +899,8 @@ mod next_boot_patch_tests {
         manager.add_patch_for_test(&temp_dir, 1)?;
         manager.record_boot_start_for_patch(1)?;
 
-        // Simulate a restart.
+        // Simulate that the app is being started fresh (e.g. from a crash)
+        manager = PatchManager::manager_for_test(&temp_dir);
         manager.on_init()?;
 
         assert!(manager.next_boot_patch().is_none());
@@ -921,7 +922,8 @@ mod next_boot_patch_tests {
         manager.add_patch_for_test(&temp_dir, 2)?;
         manager.record_boot_start_for_patch(2)?;
 
-        // Simulate a restart.
+        // Simulate that the app is being started fresh (e.g. from a crash)
+        manager = PatchManager::manager_for_test(&temp_dir);
         manager.on_init()?;
 
         assert!(manager
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -277,9 +277,9 @@ impl PatchManager {
     fn is_currently_booting_patch(&self, patch_number: usize) -> bool {
         if let Some(currently_booting_patch) = &self.patches_state.currently_booting_patch {
             return currently_booting_patch.number == patch_number;
-        } else {
-            return false;
         }
+
+        return false;
     }
 
     /// The number of the patch we last successfully booted, if any.
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -582,6 +582,42 @@ mod debug_tests {
 }
 
 #[cfg(test)]
+mod on_init_tests {
+    use super::*;
+
+    #[test]
+    fn clears_currently_booting_patch() -> Result<()> {
+        let temp_dir = TempDir::new("patch_manager").unwrap();
+        let mut manager = PatchManager::manager_for_test(&temp_dir);
+
+        // Add a patch and start to boot from it.
+        manager.add_patch_for_test(&temp_dir, 1)?;
+        manager.record_boot_start_for_patch(1)?;
+
+        assert_eq!(
+            manager
+                .patches_state
+                .currently_booting_patch
+                .as_ref()
+                .map(|p| p.number),
+            Some(1)
+        );
+        assert_eq!(
+            manager.next_boot_patch().as_ref().map(|p| p.number),
+            Some(1)
+        );
+
+        // Simulate a restart.
+        manager.on_init()?;
+
+        // Verify that we are no longer booting from patch 1.
+        assert!(manager.next_boot_patch().is_none());
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
 mod add_patch_tests {
     use super::*;
     use std::path::Path;
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -228,6 +228,7 @@ impl PatchManager {
         // If the last boot we tried was this patch, make sure we succeeded or the patch is bad.
         // If we are currently in the process of booting this patch for the first time, this check
         // will always fail, so skip it.
+        // TODO: this validation only needs to happen once. Move to on_init.
         if !self.is_currently_booting_patch(patch.number)
             && self.is_patch_last_attempted_patch(patch.number)
         {
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -273,6 +273,7 @@ impl PatchManager {
             .unwrap_or(false)
     }
 
+    /// Whether this patch is in the process of booting.
     fn is_currently_booting_patch(&self, patch_number: usize) -> bool {
         if let Some(currently_booting_patch) = &self.patches_state.currently_booting_patch {
             return currently_booting_patch.number == patch_number;
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -369,6 +369,7 @@ impl ManagePatches for PatchManager {
     fn on_init(&mut self) -> Result<()> {
         // If we were booting a patch but never recorded a successful boot, we assume that
         // the patch failed to boot. Attempt to fall back.
+        // TODO: this should record a PatchInstallFailure event. https://github.com/shorebirdtech/updater/issues/188
         if let Some(failed_boot_patch) = self.patches_state.currently_booting_patch.clone() {
             self.try_fall_back_from_patch(failed_boot_patch.number);
             self.patches_state.currently_booting_patch = None;
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -46,6 +46,10 @@ struct PatchesState {
     /// as the last booted patch patch if no new patch has been downloaded.
     next_boot_patch: Option<PatchMetadata>,
 
+    /// The patch that is currently booting, if any. If the system initializes and
+    /// this has a value, we will consider this patch to have failed to boot.
+    currently_booting_patch: Option<PatchMetadata>,
+
     /// The highest patch number we have seen. This may be higher than the last booted
     /// patch or next patch if we downloaded a patch that failed to boot.
     highest_seen_patch_number: Option<usize>,
@@ -54,6 +58,10 @@ struct PatchesState {
 /// Abstracts the process of managing patches.
 #[cfg_attr(test, automock)]
 pub trait ManagePatches {
+    /// Triggers any initialization logic needed by the patch manager. This is intended
+    /// to be called when Shorebird is initialized by the Flutter engine.
+    fn on_init(&mut self) -> Result<()>;
+
     /// Copies the patch file at file_path to the manager's directory structure sets
     /// this patch as the next patch to boot.
     ///
@@ -213,7 +221,10 @@ impl PatchManager {
         }
 
         // If the last boot we tried was this patch, make sure we succeeded or the patch is bad.
-        if self.is_patch_last_attempted_patch(patch.number) {
+        // If we are currently booting this patch, skip this check, as
+        if !self.is_currently_booting_patch(patch.number)
+            && self.is_patch_last_attempted_patch(patch.number)
+        {
             // We are trying to boot from the same patch that we tried to boot from last time.
 
             match self.last_successful_boot_patch_number() {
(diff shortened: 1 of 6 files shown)


library/src/cache/patch_manager.rs
```diff
@@ -261,6 +273,15 @@ impl PatchManager {
             .unwrap_or(false)
     }
 
+    /// Whether this patch is in the process of booting.
+    fn is_currently_booting_patch(&self, patch_number: usize) -> bool {
+        if let Some(currently_booting_patch) = &self.patches_state.currently_booting_patch {
+            return currently_booting_patch.number == patch_number;
+        } else {
+            return false;
+        }
```
eseidel, line 282: nit
```suggestion
        }
        return false;
```

eseidel: Feel free to reject if our style says otherwise. 🤷

library/src/cache/patch_manager.rs
```diff
@@ -346,6 +367,18 @@ impl PatchManager {
 }
 
 impl ManagePatches for PatchManager {
+    fn on_init(&mut self) -> Result<()> {
+        // If we were booting a patch but never recorded a successful boot, we assume that
+        // the patch failed to boot. Attempt to fall back.
+        if let Some(failed_boot_patch) = self.patches_state.currently_booting_patch.clone() {
+            self.try_fall_back_from_patch(failed_boot_patch.number);
+            self.patches_state.currently_booting_patch = None;
+            self.save_patches_state()?;
```
eseidel, line 376: Should this just be the higher-level mark failed call?  Will this end up creating a failure event?

library/src/cache/patch_manager.rs
```diff
@@ -445,7 +478,8 @@ impl ManagePatches for PatchManager {
             );
         }
 
-        self.patches_state.last_attempted_patch = Some(next_boot_patch);
+        self.patches_state.last_attempted_patch = Some(next_boot_patch.clone());
+        self.patches_state.currently_booting_patch = Some(next_boot_patch.clone());
```
eseidel, line 482: I'm confused.  Didn't we already have the concept of a currently booting patch?  How did we successfully mark them failing on failed boot before?

library/src/updater.rs
```diff
@@ -138,6 +138,15 @@ pub fn init(
...
 
+    let _ = with_config(|config| {
+        UpdaterState::load_or_new_on_error(
+            &config.storage_dir,
+            &config.release_version,
+            config.patch_public_key.as_deref(),
+        )
+        .on_init()
+    });
+
```
eseidel: Currently set_config guards against being called twice, but this line does not.  So in the case of FCM (or similar) where there are multiple Flutter Engine instances being inited in parallel across multiple threads, we will end up calling on_init() twice, even though you probably don't want to.

We could make set_config return a bool or some other way to indicate failure and bail out.  set_config could return a Result<> and and Err in the case of already being init?  That would prevent us from reaching here.

eseidel (commented):
Discussed over VC:

library/src/updater.rs
```diff
@@ -130,22 +131,29 @@ pub fn init(
...
+        Ok(()) => {
+            let _ = with_config(|config| {
+                UpdaterState::load_or_new_on_error(
+                    &config.storage_dir,
+                    &config.release_version,
+                    config.patch_public_key.as_deref(),
+                )
+                .on_init()
+            });
+        }
```
eseidel, line 155: I would early-return in the Err case, so this block doesn't need to be nested.

library/src/cache/patch_manager.rs
```diff
@@ -817,10 +889,13 @@ mod next_boot_patch_tests {
         let temp_dir = TempDir::new("patch_manager")?;
         let mut manager = PatchManager::manager_for_test(&temp_dir);
 
-        // Add a first patch and pretend it booted successfully.
+        // Add a first patch and record that we started booting it, but not that it succeeded.
         manager.add_patch_for_test(&temp_dir, 1)?;
         manager.record_boot_start_for_patch(1)?;
 
+        // Simulate a restart.
+        manager.on_init()?;
```
eseidel, line 897: This can never happen in real life though, right?

library/src/cache/patch_manager.rs
```diff
@@ -54,6 +58,10 @@ struct PatchesState {
 /// Abstracts the process of managing patches.
 #[cfg_attr(test, automock)]
 pub trait ManagePatches {
+    /// Triggers any initialization logic needed by the patch manager. This is intended
+    /// to be called when Shorebird is initialized by the Flutter engine.
```
eseidel, line 62: Intended to be called from Updater::init(), correct?

library/src/cache/patch_manager.rs
```diff
@@ -213,7 +221,11 @@ impl PatchManager {
         }
 
         // If the last boot we tried was this patch, make sure we succeeded or the patch is bad.
-        if self.is_patch_last_attempted_patch(patch.number) {
+        // If we are currently in the process of booting this patch for the first time, this check
+        // will always fail, so skip it.
```
eseidel, line 225: Is the real fix to explicitly validate as a separate phase? If so, we should say that as a TODO or issue.

library/src/cache/patch_manager.rs
eseidel: Should this return a value so the caller can create a failure event if this cant?

library/src/cache/patch_manager.rs
eseidel: Please leave yourself a note that you intend to delete the last_attempted_patch.

library/src/cache/patch_manager.rs
```diff
@@ -817,10 +889,13 @@ mod next_boot_patch_tests {
         let temp_dir = TempDir::new("patch_manager")?;
         let mut manager = PatchManager::manager_for_test(&temp_dir);
 
-        // Add a first patch and pretend it booted successfully.
+        // Add a first patch and record that we started booting it, but not that it succeeded.
         manager.add_patch_for_test(&temp_dir, 1)?;
         manager.record_boot_start_for_patch(1)?;
 
+        // Simulate a restart.
```
eseidel, line 896: This is really simulating a second start of the app.  e.g. the app crashed and is being started again.

library/src/cache/patch_manager.rs
```diff
@@ -836,9 +911,13 @@ mod next_boot_patch_tests {
         manager.record_boot_start_for_patch(1)?;
         manager.record_boot_success()?;
 
+        // Add a second patch and record that we started booting it, but not that it succeeded.
         manager.add_patch_for_test(&temp_dir, 2)?;
         manager.record_boot_start_for_patch(2)?;
 
+        // Simulate a restart.
```
eseidel, line 918: // Simulate that the app is being started fresh (e.g. from a crash)
// not bothering to re-load the state from disk.

library/src/config.rs
```diff
@@ -235,30 +239,37 @@ mod tests {
...
         );
+
+        Ok(())
     }
 
     // These tests are serial because they modify global state.
     #[serial]
     #[test]
-    fn set_config_is_noop_on_subsequent_calls() {
+    fn set_config_returns_err_on_subsequent_calls() -> Result<()> {
```
eseidel, line 249: Do we also need to test that on_init isn't called the second time?

eseidel (approved):
Needs another round, but I'm generally OK.

library/src/cache/patch_manager.rs
bryanoltman: Made fancy with:

```rust
match &self.patches_state.currently_booting_patch {
  Some(currently_booting_patch) => currently_booting_patch.number == patch_number,
  None => false,
}
```

library/src/updater.rs
bryanoltman: Fixed

library/src/updater.rs
bryanoltman: Done

library/src/cache/patch_manager.rs
```diff
@@ -46,6 +46,10 @@ struct PatchesState {
     /// as the last booted patch patch if no new patch has been downloaded.
     next_boot_patch: Option<PatchMetadata>,
 
+    /// The patch that is currently booting, if any. If the system initializes and
+    /// this has a value, we will consider this patch to have failed to boot.
```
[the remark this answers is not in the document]
bryanoltman: Done

library/src/cache/patch_manager.rs
bryanoltman: Added a note

library/src/cache/patch_manager.rs
bryanoltman: Added a todo, this goes away in the next PR that's on deck and approved.

library/src/cache/patch_manager.rs
bryanoltman: Added a TODO referencing https://github.com/shorebirdtech/updater/issues/188.

library/src/cache/patch_manager.rs
bryanoltman: Discussed on VC, this is going away in https://github.com/shorebirdtech/updater/pull/186

library/src/cache/patch_manager.rs
bryanoltman: Updated test (and other similar tests) to be clearer

library/src/cache/patch_manager.rs
bryanoltman: Updated to create a new patch manager to better simulate a fresh app start

library/src/cache/patch_manager.rs
bryanoltman: Same as above

library/src/cache/patch_manager.rs
```diff
@@ -346,6 +372,19 @@ impl PatchManager {
 }
 
 impl ManagePatches for PatchManager {
+    fn on_init(&mut self) -> Result<()> {
+        // If we were booting a patch but never recorded a successful boot, we assume that
+        // the patch failed to boot. Attempt to fall back.
+        // TODO: this should record a PatchInstallFailure event. https://github.com/shorebirdtech/updater/issues/188
+        if let Some(failed_boot_patch) = self.patches_state.currently_booting_patch.clone() {
```
eseidel, line 379: Would our previous logic have recorded an event?

library/src/cache/patch_manager.rs
```diff
@@ -538,13 +580,50 @@ mod debug_tests {
...
+            manager.next_boot_patch().as_ref().map(|p| p.number),
+            Some(1)
+        );
+
+        // Simulate that the app is being started fresh (e.g. from a crash)
+        manager = PatchManager::manager_for_test(&temp_dir);
+        manager.on_init()?;
+
+        // Verify that we are no longer booting from patch 1.
+        assert!(manager.next_boot_patch().is_none());
```
eseidel, line 621: Can you add a test that it was Some() before on_init?

library/src/cache/patch_manager.rs
bryanoltman: Nope. This isn't a regression, so I decided to fix it in a followup PR.

library/src/cache/patch_manager.rs
bryanoltman: We already test for this. Immediately above the reinitialization of the patch manager, we verify that next_boot_patch is numbered `1`.

eseidel (commented):
My only concern is if there is a regression in reporting as a result of this (e.g. if we released this would any crashes of apps start just being silent to our analytics).

Also I'm not clear as to what InitError (as a separate type) adds over just ahving anohter enum entry for UpdateError 🤷   I don't think it matters much one way or another, but you might have saved yourself some work by adding a new enum value rather tahn a whole new type.

library/src/cache/patch_manager.rs
eseidel: Oh wow.  So how does failure reporting ever work then?  I guess only failures where the failure type is "can't parse this snapshot" or similar are reported and any "crash during dart init" are just invisible to users?

library/src/cache/patch_manager.rs
eseidel: True.  I guess I would have tested that *after* deserialization, before init to test that it's actually being serialized.  🤷

bryanoltman: Ah! Understood. Done.

library/src/cache/patch_manager.rs
bryanoltman: We submit events for failures that are reported via the `report_launch_failure`, which comes in through the Flutter engine. But yeah, we're potentially dropping a lot of failure events here.

PR #189, opened 2024-07-17
fix: report launch failure when a patch was in the process of booting on app start

bryanoltman:
## Description

Moves the logic that checks whether a patch crashed while booting to the Updater level, and reports a PatchInstallFailure event if a crashed patch is detected. Patches that crashed while booting are cleaned up and will not be booted from again.

Fixes https://github.com/shorebirdtech/updater/issues/188

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -72,14 +72,6 @@ pub trait ManagePatches {
         signature: Option<&'a str>,
     ) -> Result<()>;
 
-    /// Uninstalls the artifacts for the patch with the given number.
-    ///
-    /// If the patch is the next boot patch, it will be replaced by the last booted patch if the
-    ///   last booted patch passes validation, or None if it does not.
-    /// If the patch is the last booted patch, it will be removed and the last booted patch will be
-    ///   cleared.
-    fn remove_patch(&mut self, patch_number: usize) -> Result<()>;
-
     /// Returns the patch we most recently successfully booted from (usually the currently running patch),
     /// or None if no patch is installed.
     fn last_successfully_booted_patch(&self) -> Option<PatchInfo>;
@@ -368,13 +360,6 @@ impl ManagePatches for PatchManager {
             .map(|highest_patch_number: usize| highest_patch_number.max(patch_number))
             .or(Some(patch_number));
         self.save_patches_state()
-    }
-
-    fn remove_patch(&mut self, patch_number: usize) -> Result<()> {
-        // Call try_fall_back_from_patch to ensure that we handle the case where we're removing
-        // the next_boot_patch.
-        self.try_fall_back_from_patch(patch_number);
-        Ok(())
     }
 
     fn last_successfully_booted_patch(&self) -> Option<PatchInfo> {
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -55,7 +55,11 @@ struct PatchesState {
     highest_seen_patch_number: Option<usize>,
 }
 
-/// Abstracts the process of managing patches.
+/// Abstracts the storage of patches on disk.
+///
+/// The impementation of this (PatchManager) should only be responsible for translating what is on
+/// disk into a form that is useful for the updater and vice versa. Some business logic has crept in
+/// in the form of validation, and we should consider moving that into a separate module.
 #[cfg_attr(test, automock)]
 pub trait ManagePatches {
     /// Copies the patch file at file_path to the manager's directory structure sets
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -58,9 +58,9 @@ struct PatchesState {
 /// Abstracts the process of managing patches.
 #[cfg_attr(test, automock)]
 pub trait ManagePatches {
-    /// Triggers any initialization logic needed by the patch manager. This is intended
-    /// to be called when Shorebird is initialized by the Flutter engine (shorebird_init).
-    fn on_init(&mut self) -> Result<()>;
+    // Triggers any initialization logic needed by the patch manager. This is intended
+    // to be called when Shorebird is initialized by the Flutter engine (shorebird_init).
+    //fn reset_boot_state(&mut self) -> Result<()>;
 
     /// Copies the patch file at file_path to the manager's directory structure sets
     /// this patch as the next patch to boot.
@@ -75,6 +75,14 @@ pub trait ManagePatches {
         hash: &str,
         signature: Option<&'a str>,
     ) -> Result<()>;
+
+    /// Uninstalls the artifacts for the patch with the given number.
+    ///
+    /// If the patch is the next boot patch, it will be replaced by the last booted patch if the
+    ///   last booted patch passes validation, or None if it does not.
+    /// If the patch is the last booted patch, it will be removed and the last booted patch will be
+    ///   cleared.
+    fn remove_patch(&mut self, patch_number: usize) -> Result<()>;
 
     /// Returns the patch we most recently successfully booted from (usually the currently running patch),
     /// or None if no patch is installed.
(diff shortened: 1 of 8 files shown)


library/src/cache/patch_manager.rs
```diff
@@ -316,19 +314,6 @@ impl PatchManager {
...
-        // the patch failed to boot. Attempt to fall back.
-        // TODO: this should record a PatchInstallFailure event. https://github.com/shorebirdtech/updater/issues/188
-        if let Some(failed_boot_patch) = self.patches_state.currently_booting_patch.clone() {
-            self.try_fall_back_from_patch(failed_boot_patch.number);
-            self.patches_state.currently_booting_patch = None;
-            self.save_patches_state()?;
-        }
-
-        Ok(())
-    }
```
bryanoltman, line 330: This logic has been moved up to the updater.rs level

library/src/updater.rs
```diff
@@ -51,6 +50,7 @@ impl Display for UpdateStatus {
 pub enum InitError {
     InvalidArgument(String, String),
     AlreadyInitialized,
+    FailedToCleanUpFailedPatch,
```
eseidel, line 53: I'm not sure the contract here.  The first one is fatal (you can't continue using updater until you call init correctly), the second two are benign, right?  Is this differentiated to the C callers?

library/src/updater.rs
```diff
@@ -161,22 +178,53 @@ pub fn init(
...
+    })
 }
 
+/// Whether the auto-update flag is set to true in the config.
 pub fn should_auto_update() -> anyhow::Result<bool> {
     with_config(|config| Ok(config.auto_update))
 }
 
+/// Creates a `PatchEvent` for the given `EventType` and patch number for reporting to the server.
+fn patch_event(config: &UpdateConfig, event_type: EventType, patch_number: usize) -> PatchEvent {
```
eseidel, line 216: Is this essentially a constructor/factory?  Does it need to be closer to PatchEvent?

eseidel (commented):
This is more complicated than I thought (due to the changes to updater_internal which may involve threading concerns) :). will give it another look in a minute.

library/src/updater.rs
bryanoltman: Yeah, good call. I was mimicking the `patch_check_request` function below, but it definitely makes more sense as a constructor. Done.

library/src/updater.rs
bryanoltman: Any of these being returned as an `Err` result from the `init` function results in `shorebird_init` returning false. The contract as I've been thinking about it is:

- If `shorebird_init` returns true, we can confidently say that the updater is initialized. 
- If it returns false, that particular call to init was not successful, but that doesn't necessarily mean that the updater is not ready to use (e.g., in the case of `AlreadyInitialized`). I would view the failure to clean up a failed patch as less benign than `AlreadyInitialized`, as it could mean that the bad patch is still in place as our next boot patch.

Not sure if that explanation helps. Maybe there's a better name for this or missing docs?

bryanoltman:
> due to the changes to updater_internal which may involve threading concerns

These changes should be no-ops (and I can revert if needed). The `with_state` function I introduced is just a convenience for the common pattern we had of using `with_config` and immediately constructing an `UpdaterState`. If anything, I would think this version is safer, as we are no longer accessing an `UpdaterState` instance when we don't have a lock.

library/src/updater.rs
```diff
@@ -290,43 +324,28 @@ fn update_internal(_: &UpdaterLockState) -> anyhow::Result<UpdateStatus> {
...
-    let events = read_only_state.copy_events(3);
+    let events = with_state(|state| Ok(state.copy_events(3)))?;
     for event in events {
         let result = crate::network::send_patch_event(event, &config);
         if let Err(err) = result {
             error!("Failed to report event: {:?}", err);
         }
     }
-    // We're abusing the config lock as a UpdateState lock for now.
-    let read_only_state = with_config(|_| {
```
bryanoltman, line 312: This is what I'm referring to when I say that we're no longer accessing an `UpdaterState` outside of a lock.

library/src/updater.rs
eseidel: I'm just wondering about the C++ code:
https://github.com/shorebirdtech/engine/blob/shorebird/dev/shell/common/shorebird/shorebird.cc#L128

Right now it looks like it doesn't even check the return value.  I'm just not sure what behavior it should have if it did.
Should it continue to call shorebird_ methods after a false return?  Does it need to check some other value in that case?  e.g. consider the case of the FCM multi-threaded call.  When the second thread calls shorebird_init and gets back false (currently ignored).  It presumably should keep calling shorebird methods?  But if any thread ever got the "failed to cleanup failed patch" response, is it allowed to keep calling shorebird methods?  Certainly in the "invalid argument" case, it seems like calling more shorebird methods is folly (although I suspect that since active_path returns null, after a failed init everything is fine?)
https://github.com/shorebirdtech/engine/blob/shorebird/dev/shell/common/shorebird/shorebird.cc#L143

eseidel: Maybe the answer is that shorebird_init shouldn't even return a bool and the contract is "its always OK to keep callign shorebird methods, they'll never crash, just they might also not do anything"?

bryanoltman: > Maybe the answer is that shorebird_init shouldn't even return a bool and the contract is "its always OK to keep callign shorebird methods, they'll never crash, just they might also not do anything"?

I think this is the answer. We have to support this for the `shorebird_code_push` package, and it would be nice to have the same set of assumptions for interacting with the engine. Changing the return value of the `shorebird_init` C API seems like a separate change if that works for you.

eseidel: sg.  We should document this assumption then (either as part of this change or the other), that these error conditions are just advisory (they just log) and that all shorebird apis are expected to never crash/do wrong things, even after init errors.

bryanoltman: Done.

library/src/updater.rs
```diff
@@ -161,18 +177,36 @@ pub fn init(
...
 
-    let _ = with_config(|config| {
-        UpdaterState::load_or_new_on_error(
+    handle_prior_boot_failure_if_necessary()
+}
+
+/// If, at initialization time, we detect that we were in the process of booting a patch, report a
+/// failure to boot for that patch and queue an event to report the failure.
+pub fn handle_prior_boot_failure_if_necessary() -> Result<(), InitError> {
+    with_config(|config| {
```
eseidel, line 186: Should this be using your with_state?

library/src/updater.rs
```diff
@@ -290,43 +324,28 @@ fn update_internal(_: &UpdaterLockState) -> anyhow::Result<UpdateStatus> {
...
-    let read_only_state = UpdaterState::load_or_new_on_error(
-        &config.storage_dir,
-        &config.release_version,
-        config.patch_public_key.as_deref(),
-    );
 
     // We discard any events if we have more than 3 queued to make sure
     // we don't stall the client.
-    let events = read_only_state.copy_events(3);
+    let events = with_state(|state| Ok(state.copy_events(3)))?;
```
eseidel, line 330: Previously this didn't take the config lock and now it does.  Is that OK?

library/src/updater.rs
bryanoltman: That'd be nice, but we need the `config` to create the PatchEvent later on

library/src/updater.rs
```diff
@@ -128,6 +130,20 @@ fn libapp_path_from_settings(original_libapp_paths: &[String]) -> Result<PathBuf
     first.map(PathBuf::from)
 }
 
+pub fn with_state<F, R>(f: F) -> anyhow::Result<R>
```
eseidel, line 133: This is always mut.  Some callers don't need/want the mut version.  I wonder if there should be a with_state and with_mut_state?

library/src/cache/patch_manager.rs
```diff
@@ -316,19 +314,6 @@ impl PatchManager {
 }
 
 impl ManagePatches for PatchManager {
```
eseidel, line 316: If you're making this a reading/writing class (or intend it to be), I would leave yourself a comment to explain that goal.

eseidel (approved):
Please split with_state and with_mut_state.

library/src/updater.rs
bryanoltman: Split these out

PR #191, opened 2024-07-17
feat: track known_bad_patches instead of highest_seen_patch to support rollbacks

bryanoltman:
## Description

Updates our patch check logic to:

- Record patch numbers that fail to boot in a list of "known bad" patches.
- When making a patch check request, don't send a patch number. Instead, if there is a patch available for download, check whether we currently have the patch installed and ignore the new patch if so.
- If the server responds with a patch number that we have marked bad, do not attempt to download or install it.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -2,7 +2,10 @@ use super::{disk_io, signing, PatchInfo};
 use anyhow::{bail, Context, Result};
 use core::fmt::Debug;
 use serde::{Deserialize, Serialize};
-use std::path::{Path, PathBuf};
+use std::{
+    collections::HashSet,
+    path::{Path, PathBuf},
+};
 
 #[cfg(test)]
 use mockall::automock;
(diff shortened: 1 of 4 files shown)


bryanoltman:
I looks like the missing coverage is the "UpdateIsBadPatch" display string, fwiw

eseidel (commented):
Please write a design doc about what patch_check v2 should look like and lets sort that out there.

bryanoltman:
Updated the [design doc](https://docs.google.com/document/d/1bXFHAT9Y9drOup0AM_ztIuy7t7Qx-L0XZBswzZdvMdg/edit#heading=h.gs0oylh3izhn) to clarify that the behavior of the patch check endpoint (specifically around returning a patch vs. not) will not change as part of the rollbacks feature.

library/src/cache/patch_manager.rs
```diff
@@ -50,9 +53,10 @@ struct PatchesState {
     ///  - the system initializes (on_init, we take this to mean the patch failed to boot)
     currently_booting_patch: Option<PatchMetadata>,
 
-    /// The highest patch number we have seen. This may be higher than the last booted
-    /// patch or next patch if we downloaded a patch that failed to boot.
-    highest_seen_patch_number: Option<usize>,
+    /// A list of patch numbers that we have tried and failed to install.
+    /// We should never attempt to download or install these again for the
+    /// current release.
+    known_bad_patches: HashSet<usize>,
```
eseidel, line 59: Does a HashSet serialize via serde?

eseidel (approved):

library/src/cache/patch_manager.rs
bryanoltman: Yep!

PR #194, opened 2024-07-22
feat: add rollback support

bryanoltman:
## Description

Adds rollback support.

If the response to a patch check request tells us there are rolled back patches, delete them and ensure we don't try to boot from them in the future.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/test_utils.rs
+++ b/library/src/test_utils.rs
@@ -35,9 +35,8 @@ pub fn install_fake_patch(patch_number: usize) -> anyhow::Result<()> {
 pub fn write_fake_apk(apk_path: &str, libapp_contents: &[u8]) {
     use std::io::Write;
     let mut zip = zip::ZipWriter::new(std::fs::File::create(apk_path).unwrap());
-    let options = zip::write::FileOptions::default()
-        .compression_method(zip::CompressionMethod::Stored)
-        .unix_permissions(0o755);
+    let options =
+        zip::write::FileOptions::default().compression_method(zip::CompressionMethod::Stored);
     let app_path = crate::android::get_relative_lib_path("libapp.so");
     zip.start_file(app_path.to_str().unwrap(), options).unwrap();
     zip.write_all(libapp_contents).unwrap();
--- a/library/src/test_utils.rs
+++ b/library/src/test_utils.rs
@@ -1,0 +1,42 @@
+use std::fs;
+
+use crate::{
+    cache::{PatchInfo, UpdaterState},
+    config::with_config,
+};
+
+pub fn install_fake_patch(patch_number: usize) -> anyhow::Result<()> {
+    with_config(|config| {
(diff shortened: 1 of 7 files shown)


library/src/updater.rs
```diff
@@ -360,6 +360,21 @@ fn update_internal(_: &UpdaterLockState) -> anyhow::Result<UpdateStatus> {
     // Check for update.
     let patch_check_request_fn = &(config.network_hooks.patch_check_request_fn);
     let response = patch_check_request_fn(&patches_check_url(&config.base_url), request)?;
+
+    with_mut_state(|state| {
+        if let Some(rolled_back_patches) = response.rolled_back_patch_numbers {
+            if !rolled_back_patches.is_empty() {
+                info!("Rolled back patches: {:?}", rolled_back_patches);
+                for patch_number in rolled_back_patches {
+                    info!("Attempting uninstall of patch {}...", patch_number);
```
eseidel, line 369: Do you want to log this on every boot?

library/src/test_utils.rs
```diff
@@ -0,0 +1,45 @@
...
+    })
+}
+
+/// Creates a fake APK at `apk_path` and writes `libapp_contents` to its relative `libapp.so` path.
+pub fn write_fake_apk(apk_path: &str, libapp_contents: &[u8]) {
+    use std::io::Write;
+    let mut zip = zip::ZipWriter::new(std::fs::File::create(apk_path).unwrap());
+    let options = zip::write::FileOptions::default()
+        .compression_method(zip::CompressionMethod::Stored)
+        .unix_permissions(0o755);
```
eseidel, line 40: is 0 correct?  Or do you want x?

eseidel: I'm also not sure why the fake apk needs executable permissions?

eseidel (approved):

library/src/updater.rs
bryanoltman: *every boot where the user has at least one rolled back patch.

I think it might be better to just log the patch check response in its entirety though. Updated, lmk what you think

library/src/test_utils.rs
bryanoltman: Me neither tbh—this function was just moved from a [much earlier commit](https://github.com/shorebirdtech/updater/commit/226df7d08dd7899e5cad6b597f2405708e8560d4). 

According to the `unix_permissions` docs, they seem to want octal:

```
    /// The format is represented with unix-style permissions.
    /// The default is `0o644`, which represents `rw-r--r--` for files,
    /// and `0o755`, which represents `rwxr-xr-x` for directories
```

It looks like we might've just used the directory default, possibly through copilot? The tests pass without specifying a permission, so I'm just going to remove this line.

Issue #131, opened 2024-02-28
Install events are reported with a lag


eseidel:
We report install events on the next patch check, which is typically the next launch of the app.  But we use the date of that launch, rather than when the actual install occurred.

This can create confusion for users of the console:
https://github.com/shorebirdtech/shorebird/issues/1753#issuecomment-1968406611

I suspect we should include some sort of date in the event:
https://github.com/shorebirdtech/updater/blob/main/library/src/events.rs#L42

eseidel:
Specifically we should report events with the *device* time, rather than showing them in our console with the *server* received time.  That will cause spurious events when devices have bad clocks, but it's better than the current data.

eseidel:
See also https://github.com/shorebirdtech/shorebird/issues/2139

tristdrum:
Hey @eseidel! Instead of tracking users device time, wouldn't it make more sense to continue to track the server time when the users device installed the patch, and then simply showing the graph in the admin users time zone?

Otherwise, if we don't take into account timezones properly, we might see installs on the graph that happened hours apart appear right next to each other (e.g. 9am in Texas vs 9am in New York)

Also, here in Africa assuming a users device has the correct time is far from a guarantee 😂

Love your work! Thank you for doing this all in the public! Hope to contribute and be of value to you!

Hoping to implement in the next week. Been following for months! Very exciting times.

bryanoltman:
An update on this, since it's been in the "todo" lane for a bit:

What we currently report is when a patch was marked as "good" (i.e., when we were able to successfully boot from it). The time of this event is what we use to determine which billing period a patch should apply to (we only want to bill customers for successful patch installs). This event is sent to the server when the boot is successful (there is no delay between a patch successfully booting and this event being reported).

I believe most of the customer confusion comes from the fact that this is not reporting a _download_ event. A common case is that a user will deactivate a patch and then be confused when patch installs are still being reported. What is happening in these cases is that the patch was downloaded prior to deactivation, but was only successfully booted from when the user restarted the app after the patch was deactivated.

What we should probably do is to start reporting download events separately from install events. This would provide us (Shorebird) with useful information (how many users get patches and never install them?) as well as making this distinction clearer to users.

eseidel:
I think it's slightly worse than that, there are 3 times:
1. download & install
2. boot
3. event actually reaches our servers (since internet might have been off during boot)

[...]

PR #195, opened 2024-07-22
feat: report patch downloads

bryanoltman:
## Description

This PR adds a `PatchDownloaded` event that is sent at the end of an `update` when a patch is downloaded and successfully applied to the release.

Users currently have no way of knowing whether users are receiving their patch. The closest piece of information they have to this is our reporting of patch installs, which only happen when a patch is successfully booted from. This delay has caused several customers to think our "revert patch" functionality was broken, and generally does not present them with as full of a picture as we can.

Part of https://github.com/shorebirdtech/updater/issues/131

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


eseidel (approved):
I have no objection to this, i'm just also not sure of it's utility.  Since we don't have any "patch session id" or any other way to associate events within a patch we won't have any way to tie these download events to install events.

I guess one thing we can do is look at the data in aggregate and guess at how many patches get downloaded but never installed?

bryanoltman:
@eseidel the main goal here is to be able to surface downloads vs installs to users. As-is, we won't be able to tie a specific download event to a specific install event, but I was imagining that we'd create a second chart in the console to allow users to compare downloads vs installs, and to help clarify the difference (after all of this is rolled out, we shouldn't get any more "I disabled/rolled back a patch but users are still getting it" reports).

PR #202, opened 2024-07-25
fix: update current_boot_patch to always return currently running patch

bryanoltman:
## Description

Somewhere along the way in the PatchManager refactor, `current_boot_patch` was changed to return the last successfully booted patch. There existed a brief window between launch and report boot success/failure where `current_boot_patch` would not return the actual running patch, but the last successfully booted patch instead.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/updater_state.rs
+++ b/library/src/cache/updater_state.rs
@@ -171,12 +171,20 @@ impl UpdaterState {
         self.patch_manager.currently_booting_patch()
     }
 
-    /// This is the current patch that is running.
+    /// The last patch that was successfully booted (e.g., for which we record_boot_success was
+    /// called).
     /// Will be None if:
     /// - There was no good patch at time of boot.
     /// - The updater has been initialized but no boot recorded yet.
-    pub fn current_boot_patch(&self) -> Option<PatchInfo> {
+    pub fn last_successfully_booted_patch(&self) -> Option<PatchInfo> {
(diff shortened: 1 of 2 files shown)


library/src/cache/updater_state.rs
```diff
@@ -171,14 +171,22 @@ impl UpdaterState {
...
     /// Will be None if:
     /// - There was no good patch at time of boot.
     /// - The updater has been initialized but no boot recorded yet.
-    pub fn current_boot_patch(&self) -> Option<PatchInfo> {
+    pub fn last_successfully_booted_patch(&self) -> Option<PatchInfo> {
         self.patch_manager.last_successfully_booted_patch()
     }
 
+    /// This is the current patch that is running.
+    pub fn current_boot_patch(&self) -> Option<PatchInfo> {
```
eseidel, line 184: Tests I might write (at the C level), and I'd expect:
1. call current_boot_patch before init (None)
2. call after init before start (None)
3. call after start before success (patch)
4. call after success. (patch)
5. Maybe call after failure instead? (patch?)

eseidel (approved):

library/src/cache/updater_state.rs
bryanoltman: Done

PR #210, opened 2024-08-30
feat: add message to patch event

bryanoltman:
## Description

Adds an optional `message` field to the PatchEvent object.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/events.rs
+++ b/library/src/events.rs
@@ -69,11 +69,19 @@ pub struct PatchEvent {
 
     /// When this event occurred as a Unix epoch timestamp in seconds.
     pub timestamp: u64,
+
+    /// An optional message to be sent with the event.
+    pub message: Option<String>,
 }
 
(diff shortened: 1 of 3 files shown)


library/src/events.rs
```diff
@@ -69,11 +69,19 @@ pub struct PatchEvent {
 
     /// When this event occurred as a Unix epoch timestamp in seconds.
     pub timestamp: u64,
+
+    /// An optional message to be sent with the event.
+    pub message: Option<String>,
```
eseidel, line 74: Leave yourself a comment about the possible contents here, and that it's never PII.

eseidel (approved):

library/src/events.rs
bryanoltman: Done

PR #221, opened 2024-10-16
feat: prepend all log messages with "[shorebird]"

bryanoltman:
## Description

Adds `shorebird_info` etc macros to wrap logging macros from the `log` crate. These have the same function signatures as their `log` counterparts, but prepend `[shorebird]` for easier grepping.

Sample logs:

```
10-16 18:03:04.383  9690  9712 I flutter : updater::updater: [shorebird] Patch check response: PatchCheckResponse { patch_available: true, patch: Some(Patch { number: 1, hash: "92841966822f30cb9ae46a896e364ea2ee388403d93a22d275216e5c4305ffa9", download_url: "https://cdn.shorebird.cloud/api/v1/patches/6b193e92-9ff9-4263-9456-43f611244eea/android/aarch64/105960/dlc.vmcode", hash_signature: None }), rolled_back_patch_numbers: Some([]) }

10-16 18:03:04.673  9690  9712 I flutter : updater::updater: [shorebird] Patch 1 successfully downloaded. It will be launched when the app next restarts.
10-16 18:03:04.673  9690  9712 I flutter : updater::updater: [shorebird] Update thread finished with status: Update installed
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/android.rs
+++ b/library/src/android.rs
@@ -5,7 +5,7 @@ use std::path::{Path, PathBuf};
 
 // <https://stackoverflow.com/questions/67087597/is-it-possible-to-use-rusts-log-info-for-tests>
 #[cfg(test)]
-use std::println as debug; // Workaround to use println! for logs.
+use std::println as shorebird_debug; // Workaround to use println! for logs.
 
 use crate::InitError;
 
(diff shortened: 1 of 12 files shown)


library/src/android.rs
```diff
@@ -5,7 +5,7 @@ use std::path::{Path, PathBuf};
 
 // <https://stackoverflow.com/questions/67087597/is-it-possible-to-use-rusts-log-info-for-tests>
 #[cfg(test)]
-use std::println as debug; // Workaround to use println! for logs.
+use std::println as shorebird_debug; // Workaround to use println! for logs.
```
eseidel, line 8: We probably should have just made our own macros at this point.  Which sometimes call println and sometimes call debug!.

eseidel: Nevermind, i see we do.  I was just confused as to why we have these same std::println imports in multiple places, couldn't our macros just do that?

eseidel (approved):

library/src/android.rs
bryanoltman: Great idea! Done.

PR #225, opened 2024-10-29
feat(shorebird_code_push): revamp Dart API

felangel:
## Description

- wip: p.o.c of an improved package:shorebird_code_push api


shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,93 @@
...
+  /// Returns null if no patch is installed.
+  final int? installedPatchNumber;
+
+  /// The patch number of the patch that has been most recently downloaded.
+  /// If no patch has been downloaded, this will be null.
+  /// See also:
+  /// * [isUpToDate] to determine whether a new update is available.
+  /// * [ShorebirdUpdater.update] to download a new patch.
+  final int? downloadedPatchNumber;
+}
```
bryanoltman, line 71: Why put these in a separate state object as opposed to on the Updater itself? Is the goal to allow users to await a single Future for all properties vs awaiting each individual one?

felangel: I adjusted to decouple the state (e.g. installed patch and downloaded patch) from the update check. Lmk what you think?

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,93 @@
...
+
+  /// Updates the app to the latest patch.
+  /// Note: The app must be restarted for the update to take effect.
+  ///
+  /// [onDownloadProgress] is called with the number of bytes received and the
+  /// total number of bytes to be received.
+  ///
+  /// Throws an [UpdateException] if an error occurs while updating or if no new
+  /// patch is available.
+  Future<void> update({OnDownloadProgress? onDownloadProgress});
```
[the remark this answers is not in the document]
felangel: Yeah this was more of an exercise of what this could look like from the Dart side we don't have to implement it as part of this work and/or we could poll on the Dart side 🤷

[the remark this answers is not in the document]
bryanoltman: Totally fair! Just wanted to ensure you know about the limitations on the rust side.

felangel:
@eseidel @bryanoltman would love your thoughts on the proposed API. I've updated the example to use the new API and I've wired up the new `ShorebirdUpdater` to use the existing rust APIs (though things like download progress and exception propagation won't work without rust changes).

shorebird_code_push/lib/src/shorebird_updater.dart
bryanoltman: I don't have a strong opinion, it just struck me as odd that I would (as a consumer of this API) type `updater.state.installedPatch` vs `updater.installedPatch`

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,94 @@
...
+  /// {@macro updater_available_state}
+  const UpdaterAvailableState({
+    required this.installedPatchNumber,
+    required this.downloadedPatchNumber,
+  });
+
+  /// The patch number of the currently installed patch.
+  /// This is the patch that the app is currently running.
+  /// Returns null if no patch is installed.
+  final int? installedPatchNumber;
```
bryanoltman, line 58: I think I'd probably change this to "currentPatchNumber" because we print "Patch N installed" in the updater when a patch has been downloaded and inflated, but before the patch is actually running. Using installed to mean both things might be slightly confusing.

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,94 @@
...
+  factory ShorebirdUpdater() => const ShorebirdUpdaterImpl(Updater());
+
+  /// The current state of the updater which includes the currently installed
+  /// and downloaded patches.
+  Future<UpdaterState> get state;
+
+  /// Whether there are new updates available.
+  /// Returns false when there is a new patch available that has not yet been
+  /// downloaded. Otherwise, returns true.
+  Future<bool> get isUpToDate;
```
bryanoltman, line 83: This is answering "is there a patch available for download that I haven't yet downloaded?" but I worry this might be confused for "am I running the latest patch?"

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,94 @@
...
+    required this.installedPatchNumber,
+    required this.downloadedPatchNumber,
+  });
+
+  /// The patch number of the currently installed patch.
+  /// This is the patch that the app is currently running.
+  /// Returns null if no patch is installed.
+  final int? installedPatchNumber;
+
+  /// The patch number of the patch that has been most recently downloaded.
```
bryanoltman, line 60: Does this include the currently running patch?

felangel: yes happy to adjust

shorebird_code_push/lib/src/shorebird_updater.dart
felangel: I was just trying to be consistent with installed vs downloaded. Do you prefer currentPatchNumber and downloadedPatchNumber? Naming is hard 🤣

shorebird_code_push/lib/src/shorebird_updater.dart
felangel: Hmm maybe we can model this with an enum instead of a bool to be more explicit about the downloaded vs installed state 🤷

shorebird_code_push/lib/src/shorebird_updater.dart
bryanoltman: I think "current" just to avoid confusion. "runningPatchNumber" might be another option, but I personally prefer current

shorebird_code_push/lib/src/shorebird_updater.dart
bryanoltman: I like the enum idea!

- No new patch for download
- Newer patch available for download, not downloaded
- Newer patch available and downloaded, will be applied on restart

I think that covers it?

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,101 @@
...
+class Patch {
+  /// {@macro patch}
+  const Patch({required this.number});
+
+  /// The patch number.
+  final int number;
+}
+
+/// The current state of the app in terms of whether its up-to-date.
+enum UpdateState {
```
eseidel, line 55: Is this UpdateStatus?

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  /// A new update is available for download.
+  outdated,
+
+  /// The app is up to date, but a restart is required for the update to take
+  /// effect.
+  restartRequired,
+
+  /// The current status is unknown because the updater is not available.
+  unsupported,
```
eseidel, line 67: Maybe Future<UpdateStatus> just turns into Future<UpdateStatus?> instead of having "unsupported"?

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,101 @@
...
+  /// The most common reasons for this returning false are:
+  /// 1. The app is running in debug mode (Shorebird only supports release
+  ///    mode).
+  /// 2. The app was *NOT* built using `shorebird release` and does *NOT*
+  ///    contain the Shorebird engine.
+  bool get isSupported;
+
+  /// The current state of the updater which includes the currently installed
+  /// and downloaded patches.
+  Future<PatchState> get patchState;
```
eseidel, line 87: I think you can just flatten next and current into this outer object.

eseidel (commented):
Mostly I suspect you have better "Dart taste" than I do.  You could post the API to twitter or something if you wanted community feedback, but otherwise I think we can just land and iterate.  I think we need to remove the progress callback for now.

shorebird_code_push/lib/src/shorebird_updater.dart
felangel: I changed the behaviorr to throw when the engine is unavailable

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,97 @@
...
+        '''
+No update available.
+update() should only be called when checkForUpdate() returns UpdateStatus.outdated.''',
+      );
+    }
+    await _run(_updater.downloadUpdate);
+    final status = await checkForUpdate();
+    if (status != UpdateStatus.restartRequired) {
+      // TODO(felangel): surface the underlying error reason.
+      throw const UpdaterException('Failed to download update.');
```
felangel, line 94: @eseidel @bryanoltman this is where I think we need to have a way to ask the rust side what the specific error was

eseidel:
I think the way we will want to change the rust code is:

1. Make a second copy of shorebird_update at the C layer (c_api.rs)
https://github.com/shorebirdtech/updater/blob/main/library/src/c_api/mod.rs#L223
2. Make that second copy return an Error* instead of void.  Also add a shorebird_release_error(error*).

Then the rest is FFI plumbing.

library/src/c_api/mod.rs
```diff
@@ -183,6 +194,20 @@ fn path_to_c_string(path: Option<PathBuf>) -> anyhow::Result<*mut c_char> {
     })
 }
 
+fn to_update_result(status: anyhow::Result<UpdateStatus>) -> UpdateResult {
+    let result = match status {
+        Ok(status) => UpdateResult {
+            status: status as i32,
+            message: std::ptr::null(),
+        },
```
eseidel, line 202: Do you want this to return std::ptr::null() (for the whole object) when there is no update?

library/src/c_api/mod.rs
```diff
@@ -183,6 +194,20 @@ fn path_to_c_string(path: Option<PathBuf>) -> anyhow::Result<*mut c_char> {
...
+        Err(err) => UpdateResult {
+            status: SHOREBIRD_UPDATE_ERROR,
```
eseidel, line 204: We should probably make this -1 or 100 or something to avoid later colliding with status?

library/src/c_api/mod.rs
```diff
@@ -212,6 +237,16 @@ pub unsafe extern "C" fn shorebird_free_string(c_string: *mut c_char) {
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn shorebird_free_update_result(result: *mut UpdateResult) {
+    if result.is_null() {
+        return;
+    }
+    unsafe {
+        drop(Box::from_raw(result));
```
eseidel, line 246: This isn't sufficient.  We also need to free the string.

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,159 @@
...
+  Future<UpdateStatus> checkForUpdate() async {
+    if (!_isAvailable) return UpdateStatus.unavailable;
+
+    final isUpdateAvailable = await _run(_updater.checkForUpdate);
+    if (isUpdateAvailable) return UpdateStatus.outdated;
+
+    final (current, next) = await (readCurrentPatch(), readNextPatch()).wait;
+    return next != null && current?.number != next.number
+        ? UpdateStatus.restartRequired
+        : UpdateStatus.upToDate;
```
eseidel, line 87: Why is this needed?  Should _updater.checkForUpdate return more of what you need?  (I guess that woudl be another breaking chagne).

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,159 @@
...
+      case 0: // SHOREBIRD_NO_UPDATE
+        return UpdateFailureReason.noUpdate;
+      case 2: // SHOREBIRD_UPDATE_HAD_ERROR
+        return UpdateFailureReason.downloadFailed;
+      case 3: // SHOREBIRD_UPDATE_IS_BAD_PATCH
+        return UpdateFailureReason.badPatch;
+      case 4: // SHOREBIRD_UPDATE_ERROR
+        return UpdateFailureReason.unknown;
+      default:
+        return UpdateFailureReason.unknown;
```
eseidel, line 133: I think we can find a way to share these constants. I'm surprised the ffi gen didn't do it for you.

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  @override
+  Future<void> update() async {
+    if (!_isAvailable) return;
+
+    late final Pointer<UpdateResult> result;
+
+    try {
+      result = await _run(_updater.downloadUpdateWithError);
```
eseidel: I think we need to handle the case of downloadUpdateWithError not existing?

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,159 @@
...
+        return UpdateFailureReason.unknown;
+      default:
+        return UpdateFailureReason.unknown;
+    }
+  }
+}
+
+extension on UpdateFailureReason {
+  String toFailureMessage([String details = '']) {
+    switch (this) {
```
eseidel, line 140: We could provide this via a callback into Rust if you wanted. 🤷   Or have rust fill in the message for you?

shorebird_code_push/lib/src/updater.dart
```diff
@@ -28,4 +29,12 @@ class Updater {
 
   /// Downloads the latest patch, if available.
   void downloadUpdate() => bindings.shorebird_update();
+
+  /// Downloads the latest patch, if available.
+  Pointer<UpdateResult> downloadUpdateWithError() =>
+      bindings.shorebird_update_with_error();
```
eseidel: shorebird_update_with_error can be null (if we're using an old engine).

library/src/c_api/mod.rs
felangel: No because we have a status that represents no update as the failure reason.

library/src/c_api/mod.rs
felangel: yeah was wondering about that kk thanks will fix

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: Yeah it's just more breaking rust changes 🤷

library/src/c_api/mod.rs
eseidel: There are other ways we could do this if we *didn't* need to release that string.  We could change the signature to have the *Dart* create the Status object for us (and pass it in as a * or ** to update).  But then we'd still need Dart to know to call free_string.  That might be less code if you'd prefer that method.

shorebird_code_push/lib/src/shorebird_updater_io.dart
eseidel: Since neither of those hit the network I think it's fine.

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: Do we want to directly export the constants from ffigen? They aren't very dart-like so I opted to transform them to more darty enums with API docs etc.

eseidel: I don't really care one way or another.  I had just assumed that ffigen would have picked them up and thus allowing us to not have to worry about keeping two codebases in sync.

The FFI layer is already non-darty, it's an implementation detail of package:shorebird_code_push.

library/src/c_api/mod.rs
```diff
@@ -212,6 +237,17 @@ pub unsafe extern "C" fn shorebird_free_string(c_string: *mut c_char) {
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn shorebird_free_update_result(result: *mut UpdateResult) {
+    if result.is_null() {
+        return;
+    }
+    unsafe {
+        drop(CString::from_raw((*result).message as *mut c_char));
```
eseidel, line 246: You'll need to check if message is null or this will crash when it is I think?

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: I don't see a huge downside to leaving it in the Dart API since these are errors that are displayed to users consuming the Dart API and I think it makes sense to keep the decoupled (e.g the dart error messages should be more friendly/human-readable whereas the rust layer might not care and an enum is good enough).

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: The `UpdateFailureReason` is a public part of the package:shorebird_code_push API so it's not an implementation detail.

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,177 @@
...
+/// {@template shorebird_updater_io}
+/// The Shorebird IO Updater.
+/// {@endtemplate}
+class ShorebirdUpdaterImpl implements ShorebirdUpdater {
+  /// {@macro shorebird_updater_io}
+  ShorebirdUpdaterImpl(this._updater, {IsolateRun? run})
+      : _run = run ?? Isolate.run {
+    try {
+      // If the Shorebird Engine is not available, this will throw an exception.
+      _updater.currentPatchNumber();
```
eseidel, line 28: This isn't being run in an isolate, is it?  Could this hang their app?

felangel: I carried this over from the current implementation: https://github.com/shorebirdtech/updater/blob/8e7ec5a9b61a2e03b029f7ec7d245e75960b4ebf/shorebird_code_push/lib/src/shorebird_code_push_io.dart#L25

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,177 @@
...
+    await _run(_updater.downloadUpdate);
+    final (current, next) = await (readCurrentPatch(), readNextPatch()).wait;
+    final status = next != null && current?.number != next.number
+        ? UpdateStatus.restartRequired
+        : UpdateStatus.upToDate;
+    if (status == UpdateStatus.restartRequired) return;
+    throw const UpdateException(
+      message: '''
+Downloading update failed but reason is unknown due to legacy updater.
+Please upgrade the Shorebird Engine for improved error messages.''',
```
eseidel, line 135:
```suggestion
Please upgrade to Flutter 3.24.4 or later for improved error messages.''',
```

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,177 @@
...
+        return 'An unknown error occurred.';
+    }
+  }
+}
+
+/*
+
+pub const SHOREBIRD_UPDATE_IS_BAD_PATCH: i32 = 3;
+pub const SHOREBIRD_UPDATE_ERROR: i32 = 4;
+*/
```
eseidel:
```suggestion
```

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,177 @@
...
+
+      final reason = status.toFailureReason();
+      final details = result.ref.message != nullptr
+          ? result.ref.message.cast<Utf8>().toDartString()
+          : 'unknown';
+      final message = reason.toFailureMessage(details);
+      throw UpdateException(message: message, reason: reason);
+    } finally {
+      _updater.freeUpdateResult(result);
+    }
```
eseidel, line 121: Does this whole try belong nested inside the first one (e.g. with a helper function) and then the finally belong on the first try?

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  @override
+  Future<void> update() async {
+    if (!_isAvailable) return;
+
+    Pointer<UpdateResult> result = nullptr;
+
+    try {
+      result = await _run(_updater.update);
```
eseidel, line 97: If _run throws, can we end up leaking the result?

shorebird_code_push/lib/src/shorebird_updater_io.dart
eseidel: sgtm.

shorebird_code_push/lib/src/shorebird_updater_io.dart
eseidel: Great.  Yeah, none of the rust/C api is public, and it makes sense to have a translation layer.

shorebird_code_push/lib/src/shorebird_updater_io.dart
eseidel: Yeah, so I'm pretty sure that can hang your app, since this construction is done on the main isolate, and any call into the bindings might try to take the config lock and thus block on some other FlutterEngine that is currently running an update with shorebird?  @bryanoltman.

I might just leave a fixme that we may need to wrap that in an isolate call (and thus make construction of the Shorebird engine be a future?)

eseidel: https://github.com/shorebirdtech/updater/blob/main/library/src/updater.rs#L519
https://github.com/shorebirdtech/updater/blob/main/library/src/updater.rs#L143

I think we're pretty good about not holding the config across network requests.  But it looks like in at least one case we do, so this could hang:
https://github.com/shorebirdtech/updater/blob/main/library/src/updater.rs#L543
if the background engine is reporting a success or failure.

If we move all of our networking into a queue and then probably add an ASSERT that we aren't ever holding the config lock while doing networking then calls like this are probably fine.

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: I’m not sure how that would happen. If run throws then we wouldn’t have a result

shorebird_code_push/lib/src/shorebird_updater_io.dart
bryanoltman: I think being even more specific here will be useful once we know which engine rev will contain this. As it stands, if someone has already cut a release with 3.24.4, they won't be able to use this functionality even if they patch the release with the updated `package:shorebird_code_push` api.

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: What are you concerned about? The first try/catch throwing while still somehow having allocated a result? I'm not sure how that would be possible.

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: Sure happy to adjust this once we have the new engine revision

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: Added the fixme 👍

library/src/c_api/mod.rs
felangel: I updated to free the string conditionally if it's not null

library/src/c_api/mod.rs
```diff
@@ -183,6 +194,20 @@ fn path_to_c_string(path: Option<PathBuf>) -> anyhow::Result<*mut c_char> {
     })
 }
 
+fn to_update_result(status: anyhow::Result<UpdateStatus>) -> UpdateResult {
+    let result = match status {
+        Ok(status) => UpdateResult {
+            status: status as i32,
+            message: std::ptr::null(),
```
eseidel, line 201: You could do the same trick here by calling status.toString() and making a c_string.

library/src/c_api/mod.rs
```diff
@@ -212,6 +237,22 @@ pub unsafe extern "C" fn shorebird_free_string(c_string: *mut c_char) {
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn shorebird_free_update_result(result: *mut UpdateResult) {
+    if result.is_null() {
+        return;
+    }
+    unsafe {
+        // drop the message if it's not null
```
eseidel, line 246: This is good, but you can also just call shorebird_free_c_string(result.message) and save yourself some typing.

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,138 @@
...
+    if (dart.library.js_interop) './shorebird_updater_web.dart';
+import 'package:shorebird_code_push/src/updater.dart';
+
+/// The reason a call to [ShorebirdUpdater.update] failed.
+enum UpdateFailureReason {
+  /// No update is available.
+  noUpdate,
+
+  /// The update failed because the patch was invalid.
+  badPatch,
```
eseidel, line 11: I wonder if this should be patchFailed to match downloadFailed. 🤷

eseidel: Or maybe it's installFailed?

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,173 @@
...
+    return _run(
+      () {
+        try {
+          final patchNumber = _updater.currentPatchNumber();
+          return patchNumber > 0 ? Patch(number: patchNumber) : null;
+        } catch (error) {
+          throw ReadPatchException(message: '$error');
+        }
+      },
+    );
```
eseidel, line 60: I might have abstracted this into a _readPatch(_updater.currentPatchNumber) call to avoid the duplication (even if the API is still "duplicated" the internals arent).

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,173 @@
...
+  Future<UpdateStatus> checkForUpdate() async {
+    if (!_isAvailable) return UpdateStatus.unavailable;
+
+    final isUpdateAvailable = await _run(_updater.checkForUpdate);
+    if (isUpdateAvailable) return UpdateStatus.outdated;
+
+    final (current, next) = await (readCurrentPatch(), readNextPatch()).wait;
+    return next != null && current?.number != next.number
+        ? UpdateStatus.restartRequired
+        : UpdateStatus.upToDate;
```
eseidel, line 89: You do this check in two places.  Maybe it should be a shared (private) method?

shorebird_code_push/lib/src/shorebird_updater_io.dart
eseidel: I see, that's true.

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,173 @@
...
+    try {
+      if (result == nullptr) {
+        const reason = UpdateFailureReason.unknown;
+        final message = reason.toFailureMessage();
+        throw UpdateException(reason: reason, message: message);
+      }
+
+      final status = result.ref.status;
+
+      if (status == SHOREBIRD_UPDATE_INSTALLED) return;
```
eseidel, line 113: If you don't like this, you can always add extension methods onto Pointer<UpdateResult> like wasUpdateInstalled or whatever.

shorebird_code_push/lib/src/shorebird_updater_io.dart
eseidel: As it's currently written I think it's fine.  If we ever add anything else to that first try block, we could end up leaking, since really what this is trying to clean up is the allocation from within the first try block.

shorebird_code_push/lib/src/shorebird_updater.dart
felangel: good call, I like `installFailed` 👍

shorebird_code_push/README.md
```diff
@@ -34,35 +34,36 @@ this:
...
-  runApp(const MyApp());
-}
+void main() => runApp(const MyApp());
 
 // [Other code here]
 
 class _MyHomePageState extends State<MyHomePage> {
   @override
-  void initState() {
+  Future<void> initState() async {
```
eseidel, line 47: I'm not sure if async initState is kosher?

shorebird_code_push/example/lib/main.dart
```diff
@@ -225,3 +221,43 @@ class _LoadingIndicator extends StatelessWidget {
...
+class Loading<T> extends AsyncValue<T> {
+  const Loading() : super._();
+}
+
+class Loaded<T> extends AsyncValue<T> {
+  const Loaded(this.value) : super._();
+  final T value;
+}
+
+class Error<T> extends AsyncValue<T> {
```
eseidel, line 260: This is never used?

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,138 @@
...
+    if (dart.library.js_interop) './shorebird_updater_web.dart';
+import 'package:shorebird_code_push/src/updater.dart';
+
+/// The reason a call to [ShorebirdUpdater.update] failed.
+enum UpdateFailureReason {
+  /// No update is available.
+  noUpdate,
+
+  /// The update failed because the patch failed to install.
+  installFailed,
```
eseidel, line 11: Nit, I would reorder downlaod before install, since it happens before install, but 🤷

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,138 @@
...
+  final int number;
+}
+
+/// The current status of the app in terms of whether its up-to-date.
+enum UpdateStatus {
+  /// The app is up to date (e.g. running the latest patch.)
+  upToDate,
+
+  /// A new update is available for download.
+  outdated,
```
eseidel, line 79: Is this "downloadRequired"? to make it match restartRequired?  Or need(s)Download and needRestart?

shorebird_code_push/lib/src/shorebird_updater_io.dart
```diff
@@ -0,0 +1,160 @@
...
+/// The Shorebird IO Updater.
+/// {@endtemplate}
+class ShorebirdUpdaterImpl implements ShorebirdUpdater {
+  /// {@macro shorebird_updater_io}
+  ShorebirdUpdaterImpl(this._updater, {IsolateRun? run})
+      : _run = run ?? Isolate.run {
+    try {
+      // If the Shorebird Engine is not available, this will throw an exception.
+      // FIXME: Run this in an isolate or refactor the updater to avoid risking
+      // a hang.
```
eseidel, line 29:
```suggestion
      // FIXME: Run this in an isolate or refactor the updater to avoid risking
      // a hang.  If another thread is also calling into Shorebird at the same time
      // the underlying Rust code could block getting the config lock.
```

eseidel (approved):

shorebird_code_push/README.md
felangel: yeah good call changed back to using .then

shorebird_code_push/example/lib/main.dart
felangel: It is now 👍

shorebird_code_push/lib/src/shorebird_updater.dart
felangel: download isn't required ever right? Developers can choose not to download patches so I think outdated (consistent with upToDate) made more sense.

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: yup good call, done 👍

shorebird_code_push/lib/src/shorebird_updater.dart
eseidel: sg

library/src/c_api/mod.rs
```diff
@@ -46,6 +46,17 @@ pub struct AppParameters {
     pub code_cache_dir: *const libc::c_char,
 }
 
+pub const SHOREBIRD_UPDATE_ERROR: i32 = -1;
+pub const SHOREBIRD_NO_UPDATE: i32 = 0;
+pub const SHOREBIRD_UPDATE_INSTALLED: i32 = 1;
+pub const SHOREBIRD_UPDATE_HAD_ERROR: i32 = 2;
+pub const SHOREBIRD_UPDATE_IS_BAD_PATCH: i32 = 3;
```
bryanoltman, line 53: We should probably document these.

library/src/c_api/mod.rs
```diff
@@ -113,7 +124,7 @@ fn log_on_error<F, R>(f: F, context: &str, error_result: R) -> R
 where
     F: FnOnce() -> Result<R, anyhow::Error>,
 {
-    f().unwrap_or_else(|e| {
+    f().unwrap_or_else(|e: anyhow::Error| {
```
bryanoltman, line 127: I think this line can be reverted

library/src/c_api/mod.rs
```diff
@@ -183,6 +194,23 @@ fn path_to_c_string(path: Option<PathBuf>) -> anyhow::Result<*mut c_char> {
     })
 }
 
+fn to_update_result(status: anyhow::Result<UpdateStatus>) -> UpdateResult {
+    let result = match status {
+        Ok(status) => {
+            let message = status.to_string();
+            return UpdateResult {
+                status: status as i32,
+                message: allocate_c_string(message.as_str()).unwrap(),
```
bryanoltman, line 203: `unwrap` will crash if it fails. We probably want `unwrap_or_else(|| std::null_ptr)` (not 100% sure on the syntax)

library/src/c_api/mod.rs
```diff
@@ -183,6 +194,23 @@ fn path_to_c_string(path: Option<PathBuf>) -> anyhow::Result<*mut c_char> {
...
+            };
+        }
+        Err(err) => UpdateResult {
+            status: SHOREBIRD_UPDATE_ERROR,
+            message: allocate_c_string(&err.to_string()).unwrap(),
```
bryanoltman, line 208: Same here re: unwrap

shorebird_code_push/example/lib/main.dart
```diff
@@ -225,3 +224,43 @@ class _LoadingIndicator extends StatelessWidget {
     );
   }
 }
+
+sealed class AsyncValue<T> {
```
bryanoltman, line 228: This is really neat., but I worry it might make the example harder to follow. I'd either document the absolute crap out of it or remove it in favor of something more simple.

shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -0,0 +1,138 @@
...
+  // ignore: avoid_print
+  print('''
+-------------------------------------------------------------------------------
+The Shorebird Updater is unavailable in the current environment.
+-------------------------------------------------------------------------------
+This occurs when using pkg:shorebird_code_push in an app that does not
+contain the Shorebird Engine. Most commonly this is due to building with
+`flutter build` or `flutter run` instead of `shorebird release` or `shorebird preview`.
+It can also occur when running on an unsupported platform (e.g. web or desktop).
+''');
```
bryanoltman, line 59: A++ for this

felangel:
**TODOs**
- [ ] More rust tests
- [ ] Write docs
- [ ] Write blog post
- [ ] Upgrade engine
- [ ] Include engine revision in unsupported error message
- [ ] Publish pre-release on pub

shorebird_code_push/lib/src/shorebird_updater_io.dart
felangel: We shouldn’t add anything else to the first try block since the whole point of the first try is to handle exceptions when calling update 🤷

library/include/updater.h
```diff
@@ -14,6 +14,33 @@
...
+ */
+#define SHOREBIRD_UPDATE_INSTALLED 1
+
+/**
+ * An error occurred while updating. The update was not installed.
+ */
+#define SHOREBIRD_UPDATE_HAD_ERROR 2
+
+/**
+ * The downloaded patch was not installed because it was invalid.
```
bryanoltman, line 40:
```suggestion
 * A new patch is available for download but was not downloaded because it previously failed to boot.
```

bryanoltman: Just realized this is in updater.h. Will suggest this change elsewhere.

library/src/c_api/mod.rs
```diff
@@ -46,6 +46,28 @@ pub struct AppParameters {
...
+
+/// An update was installed successfully. It will boot from the update on the
+/// next app launch.
+pub const SHOREBIRD_UPDATE_INSTALLED: i32 = 1;
+
+/// An error occurred while updating. The update was not installed.
+pub const SHOREBIRD_UPDATE_HAD_ERROR: i32 = 2;
+
+/// The downloaded patch was not installed because it was invalid.
+pub const SHOREBIRD_UPDATE_IS_BAD_PATCH: i32 = 3;
```
bryanoltman, line 64:
```suggestion
/// A new patch is available for download but was not downloaded because it previously failed to boot.
pub const SHOREBIRD_UPDATE_IS_BAD_PATCH: i32 = 3;
```

library/src/c_api/mod.rs
```diff
@@ -203,12 +243,26 @@ pub extern "C" fn shorebird_next_boot_patch_path() -> *mut c_char {
...
     }
     unsafe {
-        drop(CString::from_raw(c_string));
+        drop(CString::from_raw(c_string as *mut c_char));
+    }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn shorebird_free_update_result(result: *mut UpdateResult) {
+    if result.is_null() {
```
bryanoltman, line 257:
```suggestion

/// Free an UpdateResult returned by the updater library.
#[no_mangle]
pub unsafe extern "C" fn shorebird_free_update_result(result: *mut UpdateResult) {
    if result.is_null() {
```

PR #227, opened 2024-10-30
feat: generate a rollout group number on UpdaterState creation

bryanoltman:
# WIP

## Description

Updates UpdaterState initialization to assign a "rollout group" number between 1-100 (inclusive) to the device when UpdaterState is first created, which happens whenever a new release is installed. This number is included in the patch check request and will be used on the server to determine whether the device should receive a partially rolled out patch or the previous, fully-rolled-out patch (or no patch if there is not a previous patch)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/updater_state.rs
+++ b/library/src/cache/updater_state.rs
@@ -8,8 +8,10 @@
 use std::path::{Path, PathBuf};
 
 use anyhow::Result;
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 
+use crate::config::UpdateConfig;
 use crate::events::PatchEvent;
 
 use super::patch_manager::{ManagePatches, PatchManager};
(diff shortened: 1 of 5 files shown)


library/src/cache/updater_state.rs
```diff
@@ -73,6 +82,8 @@ impl UpdaterState {
             serialized_state: SerializedState {
                 release_version,
                 queued_events: Vec::new(),
+                // Generate random number in the range [1, 100].
+                rollout_group: rand::thread_rng().gen_range(1..101),
```
eseidel, line 86: Is 1...100 the same?

library/src/network.rs
```diff
@@ -185,6 +188,7 @@ impl PatchCheckRequest {
             release_version: config.release_version.clone(),
             platform: current_platform().to_string(),
             arch: current_arch().to_string(),
+            rollout_group: UpdaterState::load_or_new_from_config(config).rollout_group(),
```
eseidel, line 191: Should this be passed a config?  Do we need to hold the config lock to call this?

eseidel (approved):

library/src/cache/updater_state.rs
eseidel: 1..=100  https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html

bryanoltman: Done

library/src/network.rs
bryanoltman: Our convention throughout the codebase has been that you get a config with `with_config`, which obtains the lock, and that anything that receives a config as a parameter assumes that the config it receives is properly locked.

PR #229, opened 2024-11-04
feat: Add support for proxy config

zwells:
## Description

Added `proxy_url` to shorebird.yaml to use with reqwest client in `network.rs` (if defined)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


library/src/network.rs
```diff
@@ -56,20 +56,39 @@ impl Default for NetworkHooks {
...
+    if let Some(proxy_url) = config.proxy_url {
+        let proxy = reqwest::Proxy::http(&proxy_url)?;
+        client_builder = client_builder.proxy(proxy);
+    }
+    
+    let client = client_builder.build()?;
+    Ok(client)
+}
+
+fn read_config() -> anyhow::Result<YamlConfig> {
```
eseidel, line 72: I think typically config is set through https://github.com/shorebirdtech/updater/blob/main/library/src/config.rs#L91

eseidel: I see, looks like the request callbacks would now need to be config-specifc. 🤔

eseidel:
I'm open to this change.  The use case of "we want to use Shorebird behind a corporate network" makes sense.

It is unfortunate that we can't (easily) fix this by using whatever existing network interface was already correctly configured.  It feels wrong to have to configure Shorebird's https client specifically.

I think the only thing blocking this landing is one of us fiddling with how to plumb the shorebird.yaml reading through to where it's normally done (and also probably change how the network callbacks are created).

zwells:
Removed the proposed yaml config var in favor of reqwest pulling the system proxy for us.

PR #244, opened 2024-11-20
chore: add cspell checking and make pass

eseidel:
Annoying sometimes, but does catch real bugs (e.g. misspelled class names)


bryanoltman (approved):
I liked my udpates, but I guess this is fine ;)

PR #259, opened 2024-12-20
feat: add windows support

bryanoltman:
## Description

Because windows works just like Android, and because Windows doesn't have to root around in the app bundle to find a libapp.so, we need to provide a platform-specific patch_base function.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/updater.rs
+++ b/library/src/updater.rs
@@ -320,7 +320,15 @@ fn patch_base(config: &UpdateConfig) -> anyhow::Result<Box<dyn ReadSeek>> {
     Ok(Box::new(base_r))
 }
 
-#[cfg(not(any(target_os = "android", test)))]
+#[cfg(all(target_os = "windows", not(test)))]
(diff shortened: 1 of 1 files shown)


library/src/updater.rs
```diff
@@ -320,7 +320,15 @@ fn patch_base(config: &UpdateConfig) -> anyhow::Result<Box<dyn ReadSeek>> {
     Ok(Box::new(base_r))
 }
 
-#[cfg(not(any(target_os = "android", test)))]
+#[cfg(all(target_os = "windows", not(test)))]
+fn patch_base(config: &UpdateConfig) -> anyhow::Result<Box<dyn ReadSeek>> {
```
eseidel, line 324: I think File already is ReadSeek, so you can just to `fs::File::open(&config.libapp_path)`

eseidel (approved):

PR #270, opened 2025-02-12
fix: roll back patches in check_for_downloadable_update

bryanoltman:
## Description

Fixes https://github.com/shorebirdtech/shorebird/issues/2883

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/updater.rs
+++ b/library/src/updater.rs
@@ -268,6 +268,10 @@ pub fn check_for_downloadable_update(channel: Option<&str>) -> anyhow::Result<bo
     let response = request_fn(&url, request)?;
     shorebird_debug!("Patch check response: {:?}", response);
 
+    if let Some(rolled_back_patches) = response.rolled_back_patch_numbers {
+        roll_back_patches_if_needed(rolled_back_patches)?;
+    }
+
     if let Some(patch) = response.patch {
         match should_install_patch(patch.number)? {
             ShouldInstallPatchCheckResult::PatchOkToInstall => Ok(true),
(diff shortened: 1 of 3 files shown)


shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -130,6 +130,9 @@ abstract class ShorebirdUpdater {
   /// Checks for available updates and returns the [UpdateStatus].
   /// This method should be used to determine the update status before calling
   /// [update].
+  /// 
+  /// If this detects that the current patch has been rolled back, the current
+  /// patch will be uninstalled.
```
eseidel, line 135:
```suggestion
  /// patch will be uninstalled.
  /// A separate call to `update()` is required to install new patches.
```

eseidel (approved):

library/src/updater.rs
```diff
@@ -268,6 +268,10 @@ pub fn check_for_downloadable_update(channel: Option<&str>) -> anyhow::Result<bo
     let response = request_fn(&url, request)?;
     shorebird_debug!("Patch check response: {:?}", response);
 
+    if let Some(rolled_back_patches) = response.rolled_back_patch_numbers {
+        roll_back_patches_if_needed(rolled_back_patches)?;
```
eseidel, line 272: Do we have any deadlock risk from this?  This is grabbing a lock now in a pathway it wasn't before?

bryanoltman: Discussed in person, but I don't think so. This should be a no-op for the standard update pathway, and check_for_downloadable_update doesn't hold on to a config lock.

Issue #255, opened 2024-12-18
feat: allow more than 3 update tracks


eseidel:
It was originally implemented as a string in our API and I suggested changing it to an enum before shipping.  I was wrong.

Customers would like more than 3 tracks, and since our backend already supports that, we just have to change the api to accept a string instead:
https://pub.dev/documentation/shorebird_code_push/latest/shorebird_code_push/UpdateTrack.html

It should be possible to stay *source* compatible since UpdateTrack.stable can just be a static/const String instead of an enum.

Abdulrasheed1729:
I would love to work on this.

Abdulrasheed1729:
Okay, thanks for the clarification.

bryanoltman:
This has been implemented.

PR #280, opened 2025-05-28
feat: add support for arbitrary track names

bryanoltman:
## Description

Updates `shorebird_code_push` to accept arbitrary strings for track names instead of the `UpdateTrack` enum.

Part of https://github.com/shorebirdtech/updater/issues/255

This will require work on the CLI and console side to support.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -20,6 +20,7 @@ words:
   - bidiff
   - bipatch
   - bootable
+  - buildlog
   - buildroot
   - bundletool
   - carryforward
@@ -61,6 +62,8 @@ words:
   - sigstore
   - staticlib
   - subosito
+  - swiftpm
+  - symbolication
   - unbootable
   - Verdana
   - vmcode


shorebird_code_push/lib/src/shorebird_updater.dart
```diff
@@ -148,17 +148,5 @@ abstract class ShorebirdUpdater {
   /// * [isAvailable], which indicates whether the updater is available.
   /// * [checkForUpdate], which should be called to check if an update is
   ///   available before calling this method.
-  Future<void> update({UpdateTrack? track});
-}
-
-/// The track to check for updates on.
-enum UpdateTrack {
```
eseidel, line 155: We can remove the breaking-ness of this change by using an extension type and making UpdateTrack a string with these extra fields.

```
extension type UpdateTrack(String value) {
  static staging = "staging";
  static beta = "beta";
  static stable = "stable";

}
```

That should make this change non-breaking.

bryanoltman: I'll need to think about naming for a bit, but making this change non-breaking will probably best be solved by introducing a new pair of methods as counterparts to `checkForUpdate` and `update` that accept a `String` track name, and update the existing implementations to simply forward `track.name` to those new mehods.

PR #284, opened 2025-07-02
fix: don't clear last successful boot patch on rollback

bryanoltman:
## Description

Fixes https://github.com/shorebirdtech/shorebird/issues/3206

WIP until I can test locally and verify this all WAI

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -465,7 +465,15 @@ impl ManagePatches for PatchManager {
     }
 
     fn remove_patch(&mut self, patch_number: usize) -> Result<()> {
-        self.try_fall_back_from_patch(patch_number)
+        self.delete_patch_artifacts(patch_number)?;
+        // If the patch to remove is the next boot patch, clear it.
+        if let Some(next_boot_patch) = self.patches_state.next_boot_patch.clone() {
+            if next_boot_patch.number == patch_number {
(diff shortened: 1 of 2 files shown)


library/src/cache/patch_manager.rs
```diff
@@ -465,7 +465,15 @@ impl ManagePatches for PatchManager {
     }
 
     fn remove_patch(&mut self, patch_number: usize) -> Result<()> {
```
eseidel, line 467: If this is only used for the rollback case, we should comment that.

library/src/cache/patch_manager.rs
```diff
@@ -465,7 +465,15 @@ impl ManagePatches for PatchManager {
...
-        self.try_fall_back_from_patch(patch_number)
+        self.delete_patch_artifacts(patch_number)?;
+        // If the patch to remove is the next boot patch, clear it.
+        if let Some(next_boot_patch) = self.patches_state.next_boot_patch.clone() {
```
eseidel, line 470: So the change here is to clear the next_boot_patch if it happens to be the same as the patch being removed.  Do we need to clear it from other fields as well?

ShadyH20:
any updates on this issue? @eseidel

PR #288, opened 2025-07-18
feat(macos): add support for unsigned binary hashing on macOS

ahmtydn:
## Description

On macOS, Shorebird applications were experiencing "This app reports version X, but the binary is different from the version X that was submitted to Shorebird" errors. This occurred because App Store builds and local builds of the same app have different code signatures, causing hash mismatches even when the actual application code was identical.

This PR implements macOS-specific binary hash comparison that removes code signatures before computing hashes, ensuring consistent hash values regardless of signing differences.

#### 1. New macOS Module (`library/src/macos.rs`)
- `create_unsigned_copy()`: Creates a temporary copy of a binary with code signature removed using `codesign --remove-signature`
- `hash_unsigned_binary()`: Computes SHA-256 hash of a binary after removing its signature
- Includes proper error handling and fallback mechanisms

#### 2. Updated Hash Verification (`library/src/updater.rs`)
- Modified `check_hash()` function to use signature-stripped hashing on macOS
- Falls back to standard hashing if signature removal fails
- Maintains existing behavior on non-macOS platforms

#### 3. Updated Patch Signature Verification (`library/src/cache/signing.rs`)
- Modified `hash_file()` function to use unsigned binary hashing on macOS
- Ensures consistency in patch validation processes

#### 4. Dependencies
- Added `tempfile = "3.8.0"` dependency for secure temporary file handling

### Technical Approach

The solution follows the same approach as signature verification tools:
1. Copy the binary to a temporary file
2. Remove the code signature using `codesign --remove-signature`
3. Compute the hash of the unsigned binary
4. Compare hashes based on actual content, not signatures

### Benefits

1. **Eliminates false positives**: App Store and local builds with identical content now produce matching hashes
2. **Improved developer experience**: Reduces confusing error messages for legitimate app updates
3. **Platform-specific optimization**: Only affects macOS while preserving existing behavior elsewhere

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

## Testing

- ✅ Builds successfully on macOS
- ✅ Maintains backwards compatibility on all platforms
- ✅ Includes unit tests for edge cases
- ✅ Graceful fallback mechanisms for non-signed binaries
- ✅ Handles `codesign` command failures appropriately

## Files Changed

- `library/src/lib.rs` - Added macOS module declaration
- `library/src/macos.rs` - New macOS-specific functionality
- `library/src/updater.rs` - Updated hash verification logic
- `library/src/cache/signing.rs` - Updated patch signature verification
- `library/Cargo.toml` - Added tempfile dependency

## Backwards Compatibility

- ✅ Maintains existing behavior on all non-macOS platforms
- ✅ Graceful fallback to standard hashing if signature removal fails
- ✅ No changes to public APIs or interfaces

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1996,6 +1996,7 @@ dependencies = [
  "sha2",
  "simple_logger",
  "tempdir",
+ "tempfile",
  "zip",
 ]
 
--- a/library/Cargo.toml
+++ b/library/Cargo.toml
@@ -49,6 +49,8 @@ serde_json = "1.0.93"
 serde_yaml = "0.9.19"
 # For computing hashes of patch files for validation.
 sha2 = "0.10.6"
+# For creating temporary files on macOS to handle signature removal
+tempfile = "3.8.0"
 # For decompressing .apk files.
 zip = { version = "0.6.4", default-features = false, features = ["deflate"] }
 
--- a/library/src/cache/signing.rs
+++ b/library/src/cache/signing.rs
@@ -5,6 +5,18 @@ use std::path::Path;
 
 /// Reads the file at `path` and returns the SHA-256 hash of its contents as a String.
 pub fn hash_file<P: AsRef<Path>>(path: P) -> Result<String> {
+    #[cfg(target_os = "macos")]
+    {
+        // On macOS, try to compute hash after removing signature for consistency
+        match crate::macos::hash_unsigned_binary(&path) {
+            Ok(hash) => return Ok(hash),
+            Err(_) => {
+                // Fall back to regular hash computation if signature removal fails
+            }
+        }
+    }
+
+    // Regular hash computation (used on non-macOS or as fallback on macOS)
     use sha2::{Digest, Sha256}; // `Digest` is needed for `Sha256::new()`;
 
     let mut file = std::fs::File::open(path)?;
--- a/library/src/lib.rs
+++ b/library/src/lib.rs
@@ -22,6 +22,9 @@ mod yaml;
 #[cfg(any(target_os = "android", test))]
 mod android;
 
+#[cfg(target_os = "macos")]
+mod macos;
+
 #[cfg(test)]
 mod test_utils;
 
--- a/library/src/macos.rs
+++ b/library/src/macos.rs
@@ -1,0 +1,101 @@
+use anyhow::{Context, Result};
+use std::fs::File;
+use std::io;
+use std::path::Path;
+use std::process::Command;
+use tempfile::NamedTempFile;
+
+/// Creates a temporary copy of a binary file with code signature removed.
+/// This is necessary on macOS because signed and unsigned binaries have different hashes
+/// even if their core content is identical.
+pub fn create_unsigned_copy<P: AsRef<Path>>(binary_path: P) -> Result<NamedTempFile> {
+    let binary_path = binary_path.as_ref();
+    
+    // Create a temporary file to store the unsigned copy
+    let temp_file = NamedTempFile::new()
+        .with_context(|| "Failed to create temporary file for unsigned binary")?;
+    
+    // Copy the original binary to the temporary file
+    std::fs::copy(binary_path, temp_file.path())
+        .with_context(|| format!("Failed to copy binary from {:?} to temporary file", binary_path))?;
+    
+    // Remove the signature from the temporary copy using codesign
+    let output = Command::new("codesign")
+        .args(["--remove-signature", temp_file.path().to_str().unwrap()])
+        .output()
+        .with_context(|| "Failed to execute codesign command")?;
+    
+    if !output.status.success() {
+        let stderr = String::from_utf8_lossy(&output.stderr);
+        anyhow::bail!("Failed to remove signature: {}", stderr);
+    }
+    
+    Ok(temp_file)
+}
+
+/// Computes the SHA-256 hash of a binary file after removing its code signature.
+/// This is used on macOS to ensure consistent hash comparisons between signed and unsigned binaries.
+pub fn hash_unsigned_binary<P: AsRef<Path>>(binary_path: P) -> Result<String> {
+    use sha2::{Digest, Sha256};
+    
+    let unsigned_copy = create_unsigned_copy(binary_path)?;
+    
+    // Hash the unsigned copy
+    let mut file = File::open(unsigned_copy.path())?;
+    let mut hasher = Sha256::new();
+    io::copy(&mut file, &mut hasher)?;
+    let hash = hasher.finalize();
+    
+    Ok(hex::encode(hash))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::io::Write;
+    use tempfile::NamedTempFile;
+
+    #[test]
+    fn test_create_unsigned_copy() -> Result<()> {
+        // Create a test binary file
+        let mut test_file = NamedTempFile::new()?;
+        test_file.write_all(b"test binary content")?;
+        
+        // Try to create an unsigned copy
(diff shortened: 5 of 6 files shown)


bryanoltman:
I think it would be more straightforward (and not require updater changes) to simply hash the signed binary as part of `shorebird release` in the CLI

ahmtydn:
> I think it would be more straightforward (and not require updater changes) to simply hash the signed binary as part of `shorebird release` in the CLI

This is not possible because Apple re-signs the binary at multiple stages. First, when you upload to TestFlight, Apple signs the app with a beta certificate, which changes the hash. Then, when the app is released on the App Store, Apple signs it again with their production certificate, causing the hash to change once more. As a result, simply hashing the signed binary during the release process in the CLI will not produce a consistent hash, since the signature will be different at each stage.

ahmtydn:
Hi @bryanoltman ,

Did you get a chance to review it? What do you think about the improvements I made in this regard?

All the best.

eseidel:
I'm confused why this would apply to just macos and not ios.   Presumably we have a different solution for this on iOS (that maybe we should just use on Mac)?  I think it's dump_blobs iirc.

eseidel:
Yeah, actually the dump_blobs code is separate, but does the same thing: ```cc
// Instead of dumping from within this executable, we could instead
// include file offsets in the json. The problem is plumbing the file
// offsets out of the AppSnapshot parser (getting them from within the)
// various elf/macho readers is non trivial, so just dumping into a single
// concatenated file from memory here is simple.
static int HandleDumpBlobs() {
  // Get the sizes for the 4 blobs and concat them and dump them.
  // Make sure we write them in the same order they're normally stored in the
  // Elf/MachO files to minimized the patch diff.
  int64_t snapshot_data_size = Dart_SnapshotDataSize(vm_snapshot_data);
  int64_t snapshot_instr_size =
      Dart_SnapshotInstrSize(vm_snapshot_instructions);
  int64_t isolate_data_size = Dart_SnapshotDataSize(vm_isolate_data);
  int64_t isolate_instr_size = Dart_SnapshotInstrSize(vm_isolate_instructions);

  // The order of these blobs changes between container formats.
  // In Mach-O, the order is:
  //   vm_snapshot_instructions
  //   vm_isolate_instructions
  //   vm_snapshot_data
  //   vm_isolate_data
  // In ELF, the order is:
  //   vm_snapshot_data
  //   vm_isolate_data
  //   vm_snapshot_instructions
  //   vm_isolate_instructions
  // Fortunately, our patch tool is smart enough to handle this and create patches
  // that are the same size regardless of the order of the blobs.
  File* file = File::Open(nullptr, out_path, File::kWriteTruncate);
  bool success = true;
  success &= file->WriteFully(vm_snapshot_data, snapshot_data_size);
  success &= file->WriteFully(vm_isolate_data, isolate_data_size);
  success &= file->WriteFully(vm_snapshot_instructions, snapshot_instr_size);
  success &= file->WriteFully(vm_isolate_instructions, isolate_instr_size);
  success &= file->Flush();
  file->Close();
  if (!success) {
    Syslog::PrintErr("Error writing snapshot blobs to %s\n", out_path);
    return kErrorExitCode;
  }

  return 0;
}
```

ahmtydn:
Hi @eseidel,

Thanks for the context about the iOS `dump_blobs` approach. You make a good point about consistency - having platform-specific solutions that achieve the same goal isn't ideal for maintainability.

I see the logic in using snapshot extraction rather than signature removal. The `dump_blobs` approach is cleaner since it focuses on the actual Dart code content rather than dealing with platform-specific binary signing quirks.

I'm willing to adapt this approach for macOS. From what I can see in the code you referenced, it extracts and concatenates the snapshot blobs in a consistent order. For macOS implementation, I assume we'd need to:

1. Hook into the same snapshot data extraction points
2. Ensure the blob ordering matches the existing logic
3. Integrate this with the current hash verification in the updater

Technical considerations for the implementation:

**Engine side**: 
- Should this extend the existing `analyze_snapshot dump_blobs` command to support macOS, or would you prefer a separate macOS-specific implementation?
- I see the iOS version handles Mach-O vs ELF blob ordering - presumably macOS would follow the Mach-O path

**Updater side**:
- Would the hash verification logic need to invoke the engine's snapshot extraction, or should this be a build-time step that generates consistent hashes upfront?
- How should this integrate with the current `check_hash()` and `hash_file()` functions - replace the platform-specific logic entirely, or add this as another fallback layer?

**Workflow integration**:
- Should this happen during `shorebird release` (CLI-side) to pre-compute consistent hashes, or remain in the updater for runtime verification?

I'm leaning toward extending the existing `dump_blobs` functionality for macOS and having the updater call into that, but wanted to confirm the architectural direction before diving in.

Would you prefer I explore this in a new PR focused on the snapshot-based approach, or update this existing one?

Let me know if this aligns with what you had in mind.

Best,
ahmtydn

ahmtydn:
Hi @eseidel,

Thanks for the context about the iOS `dump_blobs` approach. You make a good point about consistency - having platform-specific solutions that achieve the same goal isn't ideal for maintainability.

I see the logic in using snapshot extraction rather than signature removal. The `dump_blobs` approach is cleaner since it focuses on the actual Dart code content rather than dealing with platform-specific binary signing quirks.

I'm willing to adapt this approach for macOS. From what I can see in the code you referenced, it extracts and concatenates the snapshot blobs in a consistent order. For macOS implementation, I assume we'd need to:

1. Hook into the same snapshot data extraction points
2. Ensure the blob ordering matches the existing logic
3. Integrate this with the current hash verification in the updater

Technical considerations for the implementation:

**Engine side**: 
- Should this extend the existing `analyze_snapshot dump_blobs` command to support macOS, or would you prefer a separate macOS-specific implementation?
- I see the iOS version handles Mach-O vs ELF blob ordering - presumably macOS would follow the Mach-O path

**Updater side**:
- Would the hash verification logic need to invoke the engine's snapshot extraction, or should this be a build-time step that generates consistent hashes upfront?
- How should this integrate with the current `check_hash()` and `hash_file()` functions - replace the platform-specific logic entirely, or add this as another fallback layer?

**Workflow integration**:
- Should this happen during `shorebird release` (CLI-side) to pre-compute consistent hashes, or remain in the updater for runtime verification?

I'm leaning toward extending the existing `dump_blobs` functionality for macOS and having the updater call into that, but wanted to confirm the architectural direction before diving in.

Would you prefer I explore this in a new PR focused on the snapshot-based approach, or update this existing one?

Let me know if this aligns with what you had in mind.

Best

eseidel:
It's really hard to want to engage here when I'm talking to ChatGPT rather than a person.

The bug is valid.  The fix is probably not.  I think our best course of action here is just for this to wait to get to the top of our queue and then humans will fix this rather than bots.

ahmtydn:
I should've mentioned earlier — English isn't my first language, so I use AI to help clean up my writing. The ideas and work are mine, but the polished tone might have made it feel a bit off.

I ran into this exact issue in production and spent a weekend chasing it. In my case, the hash mismatch was due to App Store vs local signing, and the codesign workaround fixed it.

Your iOS solution sounds much cleaner. I'm happy to chat more about the details — maybe over Discord text chat? My spoken English isn't great, but written is easier.

eseidel:
I completely understand.  We deal with companies and developers all over the globe and it's unfair to expect them all to speak english.  We have auto-translation on Discord for just this purpose.

I will discuss with Bryan when he's back from lunch.  I apologize for my seemingly sudden pivot.

ahmtydn:
Glad we were able to sort things out! Looking forward to hearing from you. Thanks again!

PR #300, opened 2025-10-28
feat: add uuid to updater state, patch check request

bryanoltman:
## Description

Adds logic to generate a UUID for a device if one does not already exist. This UUID is used in patch check requests, and will help us determine how many devices are running a given release, regardless of whether they have installed a patch.

UUIDs persist between version upgrades.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/library/src/cache/updater_state.rs
+++ b/library/src/cache/updater_state.rs
@@ -47,7 +47,7 @@ struct SerializedState {
     /// The client ID for this device. This is assigned on the first launch of this app and persists
     /// between release versions. This is only reset when the app is uninstalled.
     /// Shorebird uses these per-install ids in order to provide you, the customer,
-    /// install-count analytics for your apps.  Storage or use of this, and any other,
+    /// install-count analytics for your apps. Storage or use of this, and any other,
     /// information is covered in our privacy policy: https://shorebird.dev/privacy/
     client_id: String,
     // Per-release state:


library/src/cache/updater_state.rs
```diff
@@ -133,7 +155,7 @@ impl UpdaterState {
                 if !is_file_not_found(&e) {
                     shorebird_info!("No existing state file found: {:#}, creating new state.", e);
                 }
-                Self::create_new_and_save(storage_dir, release_version, patch_public_key)
+                Self::create_new_and_save(storage_dir, release_version, patch_public_key, None)
```
eseidel: Not sure this even needs to take a None here?  I would assume this constructor would just always call generate_client_id instead?

library/src/cache/updater_state.rs
```diff
@@ -44,6 +44,8 @@ pub struct UpdaterState {
 /// Written out to disk as a json file at STATE_FILE_NAME.
 #[derive(Debug, Deserialize, Serialize)]
 struct SerializedState {
+    /// The client ID for this device.
```
eseidel: What's the expected lifetime of this id?  Can we document that?  It looks like as-implemented its infinite?

library/src/cache/updater_state.rs
bryanoltman: Fixed

library/src/cache/updater_state.rs
```diff
@@ -313,13 +342,48 @@ mod tests {
...
+        let state = test_state(&tmp_dir, PatchManager::manager_for_test(&tmp_dir));
+        let original_loaded = UpdaterState::load_or_new_on_error(
+            &state.cache_dir,
+            &state.serialized_state.release_version,
+            None,
+        );
+
+        let new_loaded = UpdaterState::load_or_new_on_error(&state.cache_dir, "1.0.0+2", None);
+
+        // assert!(original_loaded.serialized_state.client_id.is_some());
```
eseidel: Intended?

library/src/cache/updater_state.rs
```diff
@@ -44,6 +44,9 @@ pub struct UpdaterState {
 /// Written out to disk as a json file at STATE_FILE_NAME.
 #[derive(Debug, Deserialize, Serialize)]
 struct SerializedState {
+    /// The client ID for this device. This is assigned on the first launch of this app and persists
+    /// between release versions. This is only reset when the app is uninstalled.
```
eseidel:
```suggestion
    /// between release versions. This is only reset when the app is uninstalled.
    /// Shorebird uses these per-install ids in order to provide you, the customer,
    /// install-count analytics for your apps.  Storage or use of this, and any other,
    /// information is covered in our privacy policy: https://shorebird.dev/privacy/
```

eseidel (approved):

library/src/cache/updater_state.rs
bryanoltman: Nope! Removed.

PR #329, opened 2026-04-02
feat: plumb module_version through as a separate field

eseidel:
## Summary
- Add `module_version: Option<String>` to `YamlConfig` (parsed from shorebird.yaml)
- Propagate through `UpdateConfig` to `PatchCheckRequest`
- Sent alongside `release_version` (skipped when null via `serde(skip_serializing_if)`)

## Context
Part of https://github.com/shorebirdtech/shorebird/issues/793

Instead of overriding `release_version` with the module version, we plumb `module_version` as a separate optional field. This keeps `release_version` as the host app's version (for analytics) while `module_version` is used for patch lookup.

**Companion PRs:**
- CLI: https://github.com/shorebirdtech/shorebird/pull/3674
- Engine + Flutter tool: https://github.com/shorebirdtech/flutter/pull/121
- Server: PR in shorebirdtech/_shorebird (updater server handler)

## Test plan
- [ ] Verify module_version is parsed from shorebird.yaml when present
- [ ] Verify module_version is null when not in yaml
- [ ] Verify PatchCheckRequest includes module_version when set
- [ ] Verify PatchCheckRequest omits module_version when null


Cargo.toml
```diff
@@ -1,3 +1,9 @@
 [workspace]
 members = ["library", "patch"]
 resolver = "2"
+
+[profile.release]
+opt-level = "z"
+lto = true
+codegen-units = 1
+strip = "debuginfo"
```
eseidel: These are spurious changes. I think this PR needs to be rebased on main.

PR #343, opened 2026-04-20
feat: send current_patch_number on patch check

easymac:
## Description

- Adds `current_patch_number: Option<usize>` to `PatchCheckRequest` (serialized as `current_patch_number`, omitted when `None`).
- Populates it from `state.current_boot_patch().map(|p| p.number)` at both call sites in `updater.rs` (`check_for_downloadable_update` and the update flow inside `with_mut_state`).
- Removes the long-standing comment explaining why we *didn't* send this.

## Motivation
The patch check is the per-launch heartbeat we already collect, so reporting the currently-running patch here gives us that dimension without a new endpoint or extra device chatter.

`current_patch_number` supersedes the legacy `patch_number` field for newer updaters. We chose a new field name rather than reusing `patch_number` because the server still keys a short-circuit response off `patch_number` for pre-[#189](https://github.com/shorebirdtech/updater/pull/189) clients (~0.7% of patch-check traffic today, Flutter ≤ 3.22.2), and sending `patch_number` from newer updaters would incorrectly engage that path and drop `rolled_back_patch_numbers` from the response.

## NOTE: Privacy
This adds a new device-attached field to a request we already send on every app launch. This change does not contradict either privacy policy doc we've published.

**Do not merge until [shorebirdtech/_shorebird#2059](https://github.com/shorebirdtech/_shorebird/pull/2059) (which restores the server-side short-circuit and reads `current_patch_number`) and [shorebirdtech/shorebird#3702](https://github.com/shorebirdtech/shorebird/pull/3702) (which adds the protocol field) are deployed to prod.**

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


library/src/network.rs
```diff
@@ -247,19 +247,25 @@ pub struct PatchCheckRequest {
     /// The unique ID of this device. This is a random UUID generated by Shorebird and _not_ the
     /// device's UUID or any other identifier that has meaning outside of Shorebird.
     pub client_id: String,
-    // We specifically do not send a patch number as part of this request because we always want to
-    // know what the latest available patch is.
+    /// The number of the patch currently running on the device, if any.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub patch_number: Option<usize>,
```
bdero: This size is not deterministic across architectures. We should probably use a fixed size (like u64, or whatever is most similar to the int sizes we store in the DB) everywhere for this.

bdero (approved):

eseidel (approved):

eseidel (approved):

PR #344, opened 2026-04-21
fix: atomic state writes and surface flush errors in disk_io

eseidel:
## Summary

- `disk_io::write` now writes to a sibling `<file>.tmp` and `rename`s it into place, and explicitly unwraps the `BufWriter` so flush errors can no longer be silently discarded by `Drop`.
- Extract `serialize_and_flush` helper so the flush-error path is unit-testable.

## Why

`BufWriter`'s `Drop` impl [silently discards flush errors](https://doc.rust-lang.org/std/io/struct.BufWriter.html#impl-Drop-for-BufWriter%3CW%3E). `patches_state.json` and `state.json` comfortably fit in `BufWriter`'s 8 KB buffer, so the only flush that actually hits disk is the implicit one at drop — and on that flush, transient I/O failures (iOS Data Protection lock, `ENOSPC`, etc.) were invisible to the caller.

Cascade:

1. `disk_io::write` returns `Ok` despite a failed flush. Target file is 0 bytes on disk (File::create truncated it).
2. `save_patches_state` → `install_patch` → `update_internal` all return `Ok(UpdateStatus::UpdateInstalled)`.
3. Dart sees `SHOREBIRD_UPDATE_INSTALLED`, no exception thrown from `ShorebirdUpdater.update()`.
4. Next session: `load_patches_state` fails to deserialize the 0-byte file and falls back to `PatchesState::default()` — `next_boot_patch: None`.
5. `checkForUpdate()` → server returns patch N → `should_install_patch(N)` sees `next_boot_patch = None` → `PatchOkToInstall` → `UpdateStatus.outdated`.

This matches a customer report of `await ShorebirdUpdater.update()` succeeding without exception but `checkForUpdate()` continuing to report `updateAvailable` instead of `readyToApply`, at a steady ~0.2%/min. Unrelated to #335; this failure mode isn't caught by the `StateStorageUnavailable` path added in the pending `79f3af8` fix because the error never surfaces from `disk_io::write` in the first place.

## Changes

`library/src/cache/disk_io.rs`:
- `write()` writes to `<path>.tmp`, then `std::fs::rename` into place. `path` is never observed truncated by a concurrent or post-crash read.
- `BufWriter::into_inner()` replaces the implicit-Drop pattern, so flush I/O errors propagate as `anyhow::Error` instead of being swallowed.
- Temp file is cleaned up on serialization/flush failure.

## Test plan

- [x] `cargo test --lib` — all 223 existing tests pass.
- [x] `cargo clippy --lib --tests -- -D warnings` — clean.
- [x] New test `write_does_not_leave_temp_file_on_success` — atomic-rename cleanup.
- [x] New test `write_preserves_existing_file_on_serialization_failure` — failed writes don't clobber `path`.
- [x] New regression test `serialize_and_flush_surfaces_error_from_inner_writer` — uses a `Write` impl that errors on every call to demonstrate the bug the fix addresses. Confirmed to **fail** on the pre-fix code (temporarily reverted `into_inner()` and watched the assertion panic) and pass on the fixed code.

### DIFF
--- a/library/src/cache/disk_io.rs
+++ b/library/src/cache/disk_io.rs
@@ -203,8 +203,11 @@ mod test {
             fn write(&mut self, _buf: &[u8]) -> std::io::Result<usize> {
                 Err(std::io::Error::other("simulated flush failure"))
             }
+            // `BufWriter::into_inner` drains its buffer via the inner
+            // writer's `write`, not its `flush`, so this path is not
+            // exercised by the test. Required by the trait.
             fn flush(&mut self) -> std::io::Result<()> {
-                Err(std::io::Error::other("simulated flush failure"))
+                Ok(())
             }
         }
 
--- a/library/src/cache/disk_io.rs
+++ b/library/src/cache/disk_io.rs
@@ -3,8 +3,8 @@ use anyhow::{bail, Context};
 use serde::{de::DeserializeOwned, Serialize};
 use std::{
     fs::File,
-    io::{BufReader, BufWriter},
-    path::Path,
+    io::{BufReader, BufWriter, Write},
+    path::{Path, PathBuf},
 };
 
 pub fn write<S, P>(serializable: &S, path: &P) -> anyhow::Result<()>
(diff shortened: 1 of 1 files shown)


eseidel:
This is an attempt to fix a customer observed issue, I think it's unlikely to actually resolve their issue, but claude seems to think it's a potential real issue at scale, that sometimes the actual flush is failing and we're only learning later.

library/src/cache/disk_io.rs
```diff
@@ -24,10 +24,54 @@ where
...
+        .into_inner()
+        .map_err(|e| anyhow::Error::new(e.into_error()))?;
+    Ok(())
+}
+
+/// Returns a sibling path in the same directory with a `.tmp` suffix,
+/// e.g. `/a/b/state.json` -> `/a/b/state.json.tmp`.
+fn temp_sibling_path(path: &Path) -> PathBuf {
+    let file_name = path.file_name().and_then(|s| s.to_str()).unwrap_or("state");
+    path.with_file_name(format!("{file_name}.tmp"))
```
bdero, line 74: For temp files and tests, prefer to use unique names to avoid parallel execution collisions down the road. E.g. by adding a random suffix.

bdero (approved):
lgtm, just one test suggestion

PR #346, opened 2026-04-22
feat: enrich patch install failure messages with diagnostics

bdero:
## Summary

The `message` field in `__patch_install_failure__` events previously contained generic strings that didn't help distinguish failure causes. This is blocking investigation of high failure rates.

### Changes

**Crash recovery** (`handle_prior_boot_failure_if_necessary`):
- Now includes `elapsed_secs` — time between boot start and crash detection. Short durations (~0-5s) suggest OOM kills during boot. Long durations (hours+) suggest user force-stops or OS process reclamation.
- Now includes `file_ok` and `file_size` — whether the patch artifact is intact at recovery time, catching corruption or partial writes.
- Example: `"crash_recovery: patch 1 failed to boot (elapsed_secs=3,file_ok=true,file_size=524288)"`

**Engine failures** (`report_launch_failure`):
- Clearer prefix for filtering: `"engine_report: patch 1 failed to launch"`

**State persistence**:
- Added `boot_started_at` (Unix timestamp) to `PatchesState`. Backward compatible — older state files deserialize this as `None`, producing `elapsed_secs=unknown`.

### Context

Companion to shorebirdtech/_shorebird#2071 which adds the `message` field to the server-side `PatchEvent` model so it actually reaches BigQuery.

## Test plan

- [x] All 223 updater library tests pass
- [x] Updated `crash_during_boot_queues_failure_event` test to verify new message format
- [x] Updated `sends_queued_events_on_update` test for new engine failure message
- [x] `boot_started_at` is backward compatible (deserialized as `None` from old state files)

### DIFF
--- a/library/src/cache/patch_manager.rs
+++ b/library/src/cache/patch_manager.rs
@@ -51,6 +51,11 @@ struct PatchesState {
     ///  - the system initializes (on_init, we take this to mean the patch failed to boot)
     currently_booting_patch: Option<PatchMetadata>,
 
+    /// Unix timestamp (seconds) when the current boot attempt started.
+    /// Used to compute elapsed time in crash recovery diagnostics.
+    /// Added in 1.6.93; older state files will deserialize this as None.
+    boot_started_at: Option<u64>,
+
     /// A list of patch numbers that we have tried and failed to install.
     /// We should never attempt to download or install these again for the
     /// current release.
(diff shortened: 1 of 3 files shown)


eseidel:
I'm not sure what he boot timing data helps with?

eseidel (approved):
I think the elapsed_secs might be slightly error prone.  But I'm OK with trying.

bdero:
> I think the elapsed_secs might be slightly error prone. But I'm OK with trying.

Yeah, the elapsed time makes no sense. Just did a flat timestamp instead.

PR #347, opened 2026-04-22
feat: report diagnostic events for failed patch updates

bdero:
## Summary

Download failures, inflate errors, hash mismatches, and install failures were previously invisible — no event was sent to the server when these occurred. Users who never received a patch were undetectable in our analytics.

### Changes

- Add new `__patch_update_failure__` event type
- Extract `download_and_install_patch()` from `update_internal()` so errors can be caught and reported
- On any failure after a patch is available, queue a `PatchUpdateFailure` event with the truncated error message (max 256 chars, no PII)
- Example: `"update_failure: patch 1 failed (Download size mismatch: expected 1234 bytes, got 567)"`

The event is queued (not sent inline) so it doesn't block the update flow, and will be delivered on the next successful update check.

### Context

Companion to #346 (enriched install failure messages) and shorebirdtech/_shorebird#2071 (server-side message field). Together these three PRs provide full visibility into the patch update pipeline.

### Server-side follow-up required

After merging, the following changes are needed to make `__patch_update_failure__` events visible in BigQuery and the admin dashboard:

1. **Cloud Logging log sink**: The `patch_install_errors` sink currently only routes `__patch_install_failure__` events. Either create a new sink or update the filter:
   ```
   gcloud logging sinks update patch_install_errors \
     --project=code-push-prod \
     --log-filter='jsonPayload.message.event.type="__patch_install_failure__" OR jsonPayload.message.event.type="__patch_update_failure__"'
   ```
   Alternatively, create a dedicated `patch_update_errors` sink/dataset.

2. **Admin dashboard model** (`web/apps/admin/app/models/code_push/patch_event.ts`): Add `PatchUpdateFailure = '__patch_update_failure__'` to the `PatchEventType` enum.

3. **BigQuery queries** (`web/packages/bigquery-api/src/client.ts`): Update dashboard queries to include the new event type, or create a separate dashboard for update failures.

4. **BigQuery schema**: Add `message` column to the new table/sink (as was done for `patch_install_errors`).

## Test plan

- [x] All 227 updater library tests pass
- [x] Existing `update_fails_on_download_size_mismatch` test verifies `PatchUpdateFailure` event is queued with error detail
- [x] Added `truncate_error_message` tests for short and long messages
- [x] Added event type serialization round-trip and unknown type deserialization tests

### DIFF
--- a/library/src/updater.rs
+++ b/library/src/updater.rs
@@ -453,6 +453,51 @@ fn update_internal(_: &UpdaterLockState, channel: Option<&str>) -> anyhow::Resul
         _ => {}
     }
 
+    let patch_number = patch.number;
+    let result = download_and_install_patch(&config, &patch, action);
+
+    if let Err(ref err) = result {
+        // Queue a diagnostic event for the failed update attempt.
+        // This captures download failures, inflate errors, hash mismatches,
+        // and install failures — all of which are otherwise invisible.
+        let message = format!(
+            "update_failure: patch {} failed ({})",
+            patch_number,
+            truncate_error_message(err)
+        );
+        let queue_result = with_mut_state(|state| {
+            state.queue_event(PatchEvent::new(
+                &config,
+                EventType::PatchUpdateFailure,
+                patch_number,
+                state.client_id(),
+                Some(&message),
+            ))
+        });
+        if let Err(queue_err) = queue_result {
+            shorebird_error!("Failed to queue update failure event: {:?}", queue_err);
+        }
+    }
+
+    result
+}
+
+/// Truncates an error message to a reasonable length for the event payload.
+/// Avoids sending excessively long messages while preserving the root cause.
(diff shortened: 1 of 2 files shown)


library/src/updater.rs
```diff
@@ -224,19 +224,29 @@ pub fn handle_prior_boot_failure_if_necessary() -> Result<(), InitError> {
             config.patch_verification,
         );
         if let Some(patch) = state.currently_booting_patch() {
+            let boot_time_info = match state.boot_started_at() {
+                Some(ts) => format!("boot_started_at={ts}"),
+                None => "boot_started_at=unknown".to_string(),
```
eseidel: Why not just empty?

eseidel (approved):

library/src/updater.rs
bdero: No reason. Done.

PR #348, opened 2026-05-01
fix: current_boot_patch survives server-driven rollback

### LINKED ISSUE #3728: bug: checkForUpdate reports upToDate after patch-to-release rollback (regression of #3206)
## Summary

`shorebird_code_push.checkForUpdate()` returns `UpdateStatus.upToDate` when
the running patch is rolled back to the base release version. This is a
follow-up to #3206 — the fix shipped in shorebird_code_push 2.0.6
(shorebirdtech/updater#312) addresses some rollback cases but not the
patch-to-release case described here.

## Reproduction

1. Install patch 1 on a device, let the app boot successfully.
2. Roll back patch 1 from the dashboard (no replacement patch).
3. From the running app, call `await updater.checkForUpdate()`.

**Expected:** `UpdateStatus.restartRequired`
**Actual:** `UpdateStatus.upToDate`

The app does revert to the release version on next launch, but the running
session has no signal to prompt a restart.

Patch-to-patch rollback (e.g. rolling back patch 2 while patch 1 is still
available) is detected correctly, because the server response includes a
downloadable patch and `checkForUpdate` short-circuits to `outdated`.

## Versions

- shorebird_code_push: 2.0.6
- Flutter: 3.41.7
- Reported by a customer migrating to Shorebird.

## Root cause

The fix in shorebirdtech/updater#312 relies on Dart seeing
`current.number != null && next.number == null` after a rollback. That
holds in the test (which mocks `currentPatchNumber` to return 1), but not
in the real FFI path:

1. After successful boot, `record_boot_success` clears
   `currently_booting_patch` and sets `last_booted_patch = patch_1`.
2. `checkForUpdate` calls `check_for_downloadable_update`, which calls
   `roll_back_patches_if_needed([1])` →
   `try_fall_back_from_patch(1)` (`library/src/cache/patch_manager.rs`).
3. Because patch 1 is both `last_booted_patch` and `next_boot_patch`,
   the "both bad" branch clears **both** fields.
4. `current_boot_patch()` is defined as
   `currently_booting_patch.or(last_successfully_booted_patch)`
   (`library/src/cache/updater_state.rs`), so it now returns `None`.
5. `shorebird_current_boot_patch_number()` returns 0,
   `shorebird_next_boot_patch_number()` returns 0,
   Dart compares `null != null` → false → `upToDate`.

The original issue's note — *"After the rollback has been applied, both
currentPatch and nextPatch report null, so that might be the real issue"*
— was correct; the Dart-only fix did not address that.

## Proposed fix

Fix on the Rust side. Options:

- Don't clear `last_booted_patch` in `try_fall_back_from_patch` for the
  patch the app is currently running, so `current_boot_patch()` keeps
  reporting it until the app restarts.
- Or expose a separate FFI signal for "the patch this process is actually
  executing" that survives rollback bookkeeping.

Will follow up with a PR in shorebirdtech/updater.

eseidel:
## Summary

Fixes shorebirdtech/shorebird#3728 (customer report: patch-to-release
rollback returns `upToDate` instead of `restartRequired`). Started as
the session-scoped fix; #349 prototyped the data-model fix; we found
they're orthogonal — different real bugs — and rolled them together
here.

## The two real bugs

**Bug 1 — data model.** `last_booted_patch` was conflated. Its field
name and `record_boot_success` say it's a *historical* record ("the
patch that last successfully booted, ever"). But
`try_fall_back_from_patch` treats it as an *operational* fallback
target and clears it whenever the patch becomes invalid as a
fallback — including under server rollback, while the running
process is still using the patch. The `currently_booting.or(
last_successfully_booted_patch)` fallback in
`UpdaterState::current_boot_patch()` then surfaced the cleared field
to the FFI as 0, even though the process was still running the patch.

**Bug 2 — process state.** There was no session-scoped signal for
"what's running this run." The FFI's `current_boot_patch()` derived
it from persisted fields, which means stale state from a previous run
could surface on a release boot — e.g., after the rollback + restart,
`last_booted_patch` could keep reporting the old patch number even
though the app was on the base release.

## What this PR does

- **Fix bug 1**: in `try_fall_back_from_patch`'s "both bad" branch,
  only clear `next_boot_patch`. Leave `last_booted_patch` alone —
  that's history. The "don't fall back to this patch" intent is
  already covered by artifact deletion, `validate_patch_is_bootable`,
  and `is_known_bad_patch`. As a side-effect, `record_boot_failure_for_patch`
  also stops erasing boot history.
- **Fix bug 2**: introduce a `current_boot_patch: Option<usize>`
  field on `PatchesState`. Set by `report_launch_start` from
  whatever `next_boot_patch` was at that moment (`Some(n)` for a
  patch boot, `None` for a release boot). `UpdaterState::current_boot_patch()`
  reads it directly — no more `.or()` fallback.

After both: each field has exactly one job and a name that matches.

| Field | Role |
|---|---|
| `last_booted_patch` | Historical record. Updated only by `record_boot_success`. Used as a fallback target by `try_fall_back_from_patch`. |
| `next_boot_patch` | What the engine should load on the next launch. `None` = base release. |
| `current_boot_patch` | What this process is using. Set by `report_launch_start`. Reset on each launch (including to `None` on a release boot). |
| `currently_booting_patch` | Boot-in-progress flag. Set by `record_boot_start_for_patch`, cleared on success/failure. Used for crash detection on next init. |

## C API contract

Unchanged. `shorebird_current_boot_patch_number` returns the same
`usize` it always has — it just gets the right answer under rollback
now, and on subsequent release boots.

## Verification (at the C API level)

- `rollback_to_release_keeps_current_boot_patch` — direct regression
  test of the customer scenario. Fails on parent commit (returns 0);
  passes here (returns 1).
- `rollback_to_release_then_restart_clears_current_boot_patch` —
  proves post-restart correctness. Stale `current_boot_patch` from
  the previous run is reset to `None` by the next `report_launch_start`
  when there's no `next_boot_patch`.
- `rollback_patch_to_patch_reports_current_and_next_distinctly` —
  proves we didn't break the patch-to-patch path. After `update()`
  installs the replacement, `current=2, next=1`.
- `preserves_last_booted_patch_on_failure_but_marks_bad` (replaces
  `clears_last_booted_patch_if_it_is_the_failed_patch`) — asserts
  the data-model fix: boot history preserved, `is_known_bad_patch`
  records the failure, artifacts deleted.
- All 228 existing tests pass without modification, including every
  C API test. clippy + fmt clean.

## Why combine the two fixes

#349 alone fixed bug 1 but left bug 2 — the post-restart staleness
became an `#[ignore]`'d test there. #348 alone fixed bug 2 but left
bug 1 — TODOs documented the data-model conflation. Each PR named
the other's gap. Landing both eliminates both gaps and the model
ends up self-consistent.

## Closes / supersedes

- Closes #349 (rolled in here).

## Test plan

- [x] Regression test fails on parent commit, passes on this branch
- [x] All existing tests pass without modification
- [x] Adjacent scenarios (post-restart, patch-to-patch, boot failure) covered
- [x] `cargo test --lib` 228 pass / 0 fail
- [x] `cargo clippy --lib --tests` clean
- [x] `cargo fmt --check` clean
- [ ] Validate on a real device against the customer scenario before landing

Refs: shorebirdtech/shorebird#3728, shorebirdtech/updater#312, #270


eseidel:
Hold on. I'm going to try a different approach.

library/src/cache/patch_manager.rs
```diff
@@ -312,9 +356,22 @@ impl PatchManager {
             .unwrap_or(false);
 
         if is_bad_patch_last_booted_patch && is_bad_patch_next_boot_patch {
```
bdero, line 340: _bdero's review bot_:

**Server-driven rollback can leave a stale fallback target.**

The comment justifies preserving `last_booted_patch` here with three protections: artifact deletion, `validate_patch_is_bootable`, and `is_known_bad_patch`. Two of those don't fire on this path:

- **`is_known_bad_patch` is not set.** `remove_patch` (line 525) calls `try_fall_back_from_patch` *without* inserting into `known_bad_patches`. Only `record_boot_failure_for_patch` does that. After a server rollback of patch 1, `is_known_bad_patch(1) == false`.
- **Artifact deletion is fire-and-forget**: `let _ = self.delete_patch_artifacts(bad_patch_number);` on line 343. If it fails (FS race, permission flap, partial delete), artifacts survive.

Concrete hazard:

1. Run A: install patch 1, boot success. State: `last_booted=1, next=1`, artifacts present.
2. Server rolls back [1]. New code path: `next=None, last_booted=Some(1)`. Artifact delete attempted — assume it fails silently.
3. Server later sends patch 2. `next=Some(2), last_booted=Some(1)`.
4. Patch 2 fails to boot → `record_boot_failure_for_patch(2)` → `try_fall_back_from_patch(2)` → enters the else-if branch → validates `last_booted=patch 1` → `validate_patch_is_bootable` succeeds (artifacts present) → sets `next_boot_patch = Some(patch 1)`.
5. We boot a patch the server explicitly told us to roll back.

Pre-fix this couldn't happen because `last_booted` was cleared during the rollback. Suggested mitigation: in this BOTH branch, also insert `bad_patch_number` into `known_bad_patches`, and have the else-if fallback path consult `is_known_bad_patch` before promoting `last_booted_patch` to `next_boot_patch`. Either change alone closes the gap; both makes the contract obvious.

library/src/cache/patch_manager.rs
```diff
@@ -37,13 +37,42 @@ struct PatchMetadata {
...
 
+    /// Patch number this process is using, captured at `report_launch_start`
+    /// from whatever `next_boot_patch` was at that moment. `None` when running
+    /// the base release. Persists for the lifetime of the run; survives a
+    /// server-driven rollback of that patch (the process is still using it,
+    /// even though it won't boot from it again). Reset on the next
+    /// `report_launch_start`. Only the patch number is needed — this field
+    /// reports running state, not bootable-patch metadata.
+    #[serde(default)]
+    current_boot_patch: Option<usize>,
```
bdero: _bdero's review bot_:

**This field is described as session-scoped but is being persisted to disk.**

The doc says "Persists for the lifetime of the run … Reset on the next `report_launch_start`." That's the description of an *in-memory* field. Persisting it then writing a workaround (`set_current_boot_patch(None)` on every release `report_launch_start`) is fighting the data model. The `rollback_to_release_then_restart_clears_current_boot_patch` test exists *because* the field is persisted — it has to assert that the previous run's staleness gets clobbered on the next run.

If this lived in memory on `UpdaterState` (e.g. a `Cell<Option<usize>>` or just a plain field), set in `report_launch_start` and never serialized:

- No need for `#[serde(default)]`.
- No need for the post-restart-clears-stale test.
- No window between `init` and `report_launch_start` where `shorebird_current_boot_patch_number()` returns last run's persisted `Some(1)`.

That last point is a real gap. The new test asserts `next_boot_patch_number() == 0` before calling `report_launch_start`, but does not assert what `current_boot_patch_number()` returns in that window. After init and before `report_launch_start`, it's whatever's on disk from the previous run — for the customer scenario, that's `Some(1)`. Anything that touches the FFI in that window (Engine, a Dart isolate spinning up) sees a phantom patch.

library/src/cache/patch_manager.rs
```diff
@@ -37,13 +37,42 @@ struct PatchMetadata {
...
+
+    // TODO(eseidel): `currently_booting_patch` is redundant now that
```
bdero: _bdero's review bot_:

**Two fields now track overlapping state — exactly the anti-pattern this PR just fixed.**

The TODO acknowledges that `currently_booting_patch` is redundant with the new `current_boot_patch`. The PR description claims "each field has exactly one job and a name that matches" — but this TODO contradicts that for the boot-in-progress case. The bug being fixed was field conflation of `last_booted_patch`; landing this PR creates a new conflation between `currently_booting_patch` and `current_boot_patch`, with a follow-up to resolve it. That's how the next subtle rollback bug gets written.

Two reasonable resolutions:

- **Land the cleanup in this PR.** The TODO says it touches `handle_prior_boot_failure_if_necessary`, the boot-record helpers, and "several tests" — non-trivial but bounded, and doing it now means the data model has one moment of inconsistency, not two.
- **Keep the redundancy but rename.** `current_boot_patch` could be `running_patch` to make the distinction obvious at every call site. Mechanical diff; payoff is future readers don't need to read a TODO to know which is which.

I'd push for option 1.

library/src/cache/patch_manager.rs
```diff
@@ -445,6 +502,17 @@ impl ManagePatches for PatchManager {
...
 
+    fn current_boot_patch(&self) -> Option<PatchInfo> {
+        self.patches_state
+            .current_boot_patch
+            .map(|number| self.patch_info_for_number(number))
+    }
+
+    fn set_current_boot_patch(&mut self, patch_number: Option<usize>) -> Result<()> {
+        self.patches_state.current_boot_patch = patch_number;
+        self.save_patches_state()
```
bdero: _bdero's review bot_:

**`report_launch_start` now writes to disk on every release boot.**

Pre-fix, a release boot in `report_launch_start` did nothing — the `if let Some(next) = ...` returned early without touching state. Post-fix, `set_current_boot_patch(next.map(...))?` runs unconditionally and ends in `save_patches_state()`. So every release boot now hits disk before we know whether anything else needs persisting.

Two implications:

- **New failure mode for release boots.** If `save_patches_state()` returns Err (full disk, transient FS issue), `report_launch_start` propagates Err to the engine on a release boot — a path that previously couldn't fail there. Worth confirming Flutter Engine's contract for a non-zero return from `shorebird_report_launch_start`. If it's treated as fatal, a transient FS hiccup just bricked the cold start.
- **Avoidable I/O.** If `current_boot_patch` is moved to in-memory state per the comment on line 61, this disappears entirely.

library/src/updater.rs
```diff
@@ -810,18 +810,19 @@ pub fn current_boot_patch() -> anyhow::Result<Option<PatchInfo>> {
...
-        if let Some(next_boot_patch) = state.next_boot_patch() {
-            state.record_boot_start_for_patch(next_boot_patch.number)
-        } else {
-            Ok(())
+        let next_boot_patch = state.next_boot_patch();
+        // Capture what this run is using. None means we're booting the base
+        // release. This is the source of truth for `current_boot_patch()` and
+        // is independent of the boot-in-progress / last-successfully-booted
+        // bookkeeping.
+        state.set_current_boot_patch(next_boot_patch.as_ref().map(|p| p.number))?;
```
bdero: _bdero's review bot_:

**Split-state risk: `current_boot_patch` is set before `record_boot_start_for_patch`, with two separate disk writes.**

```rust
state.set_current_boot_patch(next.as_ref().map(|p| p.number))?;   // disk write 1
if let Some(next) = next_boot_patch {
    state.record_boot_start_for_patch(next.number)?;              // disk write 2
}
```

If the second write fails (or the process is killed between them), on-disk state has `current_boot_patch=Some(n)` but `currently_booting_patch=None`. On next init, `handle_prior_boot_failure_if_necessary` checks `currently_booting_patch` only and won't fire — a crashed boot is silently forgotten. The next call to `report_launch_start` will overwrite `current_boot_patch`, hiding the evidence.

If the field becomes in-memory per the line-61 comment (and `currently_booting_patch` is folded into `boot_started_at.is_some()` per the line-63 TODO), this collapses to a single write. As-is, prefer one combined operation that updates both fields in a single `save_patches_state()` call.

library/src/c_api/mod.rs
```diff
@@ -925,6 +926,294 @@ mod test {
...
+        let c_params = parameters(&tmp_dir, fake_libapp_path.to_str().unwrap());
+        let c_yaml = c_string("app_id: foo");
+        assert!(shorebird_init(&c_params, FileCallbacks::new(), c_yaml));
+        free_c_string(c_yaml);
+        free_parameters(c_params);
+
+        // Set up patch 2 as the running patch.
+        testing_set_network_hooks(
+            |_url, _request| {
+                let hash = "bb8f1d041a5cdc259055afe9617136799543e0a7a86f86db82f8c1fadbd8cc45";
```
bdero: _bdero's review bot_:

**Test fixture's hash doesn't match its narrative.**

This test installs "patch 2" using the hash `bb8f1d04…` — the same hash used for patch 1 in the other two tests, with the comment elsewhere reading `Generated by string_patch "hello world" "hello tests"`. The test passes because the bytes/APK are identical between the scenarios, so what it actually proves is "two installs of the same patch artifact are reported as current=2, next=1," not "patch 2 was rolled back to patch 1."

Functionally fine for the assertions made, but the comment vs. the fixture are saying different things — easy footgun for the next person editing this. Either generate a distinct fixture for patch 2, or reword the comment to be honest that the patch *number* differs but the artifact is reused for test convenience.

bdero (commented):
_bdero's review bot_:

The headline fix is right — splitting `current_boot_patch` out as its own field is the correct shape, and the customer's #3728 scenario is genuinely addressed. A few follow-on concerns that could become the next bug, posted inline. Items 1–3 are the ones I'd want resolved before merge; 4–6 are smaller.

bdero (approved):

PR #349, opened 2026-05-01
fix: don't clear last_booted_patch when its patch is rolled back (alt to #348)

### LINKED ISSUE #3728: bug: checkForUpdate reports upToDate after patch-to-release rollback (regression of #3206)

eseidel:
## Summary

Companion to #348. Both PRs fix shorebirdtech/shorebird#3728 (customer
report: patch-to-release rollback returns \`upToDate\` instead of
\`restartRequired\`) but take different approaches. This PR exists for
side-by-side comparison; we'll pick one.

## The bug under both lenses

\`last_booted_patch\` is conflated. The field name and
\`record_boot_success\` say it's a *historical* record (\"last patch that
successfully booted, ever\"), but \`try_fall_back_from_patch\` treats it
as an *operational* fallback target and clears it when the patch
becomes invalid as a fallback. Those two roles only diverge under
server rollback. With the field cleared while the process is still
running the patch, the FFI's \`currently_booting.or(last_successfully_booted_patch)\`
fallback returns 0.

## This PR's approach

Fix the conflation directly. In \`try_fall_back_from_patch\`'s \"both
bad\" branch, only clear \`next_boot_patch\`. Leave \`last_booted_patch\`
alone — that history shouldn't change because the server told us not
to use the patch next time. The \"don't fall back to this patch\"
intent is already covered:

- Artifacts are deleted at the top of the function.
- The else-if branch's \`validate_patch_is_bootable\` refuses any
  fallback to a patch with missing artifacts.
- \`is_known_bad_patch\` records boot failures explicitly.

## What works, what doesn't

| Scenario | This PR | #348 |
|---|---|---|
| In-session rollback (customer's bug) | ✅ fixed | ✅ fixed |
| Post-restart: app on release, no false \`restartRequired\` | ❌ broken | ✅ fixed |
| Boot-failure path preserves boot history | ✅ corrected | ➖ unchanged |

The post-restart gap is documented as an \`#[ignore]\`'d test in this
branch (\`rollback_to_release_then_restart_clears_current_boot_patch\`).
After rollback + restart, \`last_booted_patch\` is still \`Some(1)\` on
disk from the previous run. There's no session-scoped signal to say
\"this run is on release.\" The \`.or()\` fallback in
\`current_boot_patch()\` returns the stale 1, and Dart sees a
perpetual \`restartRequired\`.

To close that gap within Path 1, we'd need either:
- A session-scoped field for \"what's running this run\" (which is
  what PR #348 adds), or
- A clear-on-launch hack in \`report_launch_start\` when there's no
  next patch (we discussed this — it's what the user originally
  proposed and then we both agreed was a hack across functions).

So Path 1 alone is genuinely incomplete for full correctness.

## Trade-offs

**This PR:**
- + Smaller diff (~10 lines of production code).
- + Removes the conflation in the data model (the underlying bug).
- + Boot-failure history is correctly preserved (a side improvement).
- − Doesn't fix the post-restart staleness.
- − Still relies on the \`.or()\` fallback, which has its own conceptual
  weirdness (using \"last booted\" as a proxy for \"running\").

**PR #348:**
- + Fixes both in-session and post-restart correctness.
- + Each field has a single, well-scoped role.
- + No \`.or()\` fallback — \`current_boot_patch()\` reads its dedicated
  field directly.
- − Larger diff (new field, trait methods, plumbing).
- − Sidesteps the underlying conflation rather than fixing it. TODOs
  in #348 flag the unresolved \`last_booted_patch\` weirdness.

## What I'd actually do

Combine both. Take the conflation fix from this PR (path 1 — keep
\`last_booted_patch\` historical) and the session-scoped field from
PR #348. Each addresses a different real bug; they're orthogonal.
Keeping just one leaves the other unresolved.

## Test plan

- [x] Path 1 fixes the customer's reported bug at the C API level
- [x] One existing test (\`clears_last_booted_patch_if_it_is_the_failed_patch\`)
      asserts the old behavior; updated to assert new contract under
      a new name
- [x] All other 226 existing tests pass without modification
- [x] Post-restart gap is documented as \`#[ignore]\`'d test (un-ignore
      to reproduce)
- [x] \`cargo fmt --check\` and \`cargo test --lib\` clean
- [ ] Decide between this PR, #348, or both

### DIFF
--- a/library/src/c_api/mod.rs
+++ b/library/src/c_api/mod.rs
@@ -402,6 +402,7 @@ mod test {
     use crate::{
         network::{
             testing_set_network_hooks, DownloadResult, PatchCheckResponse, UNEXPECTED_DOWNLOAD,
+            UNEXPECTED_REPORT,
         },
         test_utils::write_fake_apk,
     };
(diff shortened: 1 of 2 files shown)


eseidel:
Rolled into #348. The data-model fix here (don't clear `last_booted_patch` on rollback) addresses a real bug, but in isolation it leaves the post-restart staleness gap (the `#[ignore]`'d test). Combining with #348's session-scoped `current_boot_patch` field closes both gaps cleanly — neither PR was sufficient alone. See updated description on #348.

PR #350, opened 2026-05-02
refactor: split C API into Dart and engine surfaces

eseidel:
## Summary

- Splits `library/src/c_api` into two submodules — `dart` (Dart-stable ABI consumed by `package:shorebird_code_push`) and `engine` (engine-internal, no stability guarantee). Two cbindgen configs emit `include/updater_dart.h` and `include/updater_engine.h` respectively. The combined `include/updater.h` is removed; consumers include the specific header for their use case.
- Drops two retired Dart symbols: `shorebird_update` (replaced by `shorebird_update_with_result` in the 2.0 rewrite) and `shorebird_check_for_update` (replaced by `shorebird_check_for_downloadable_update`). The package's `_legacyFallback`/`downloadUpdate` was the only path still calling `shorebird_update`, and the package's `flutter: >=3.24.5` constraint already guarantees the new symbol is present, so the fallback was unreachable.
- Repoints ffigen at `updater_dart.h` so engine-only symbols stop appearing in `updater_bindings.g.dart`.
- Bumps `shorebird_code_push` to 2.0.7.

## Why

The Dart-FFI ABI and the engine-FFI ABI have different stability requirements but were sharing one bucket. That made it ambiguous which symbols were safe to break and let ffigen bind engine-only symbols nothing in Dart calls. Splitting the headers makes the boundary explicit before any per-engine refactoring lands on the engine side.

## Follow-up (engine repo)

Coordinated change needed in https://github.com/shorebirdtech/engine:

- [ ] `shell/common/shorebird/updater.cc` includes `third_party/updater/library/include/updater.h` — switch to `updater_engine.h`.
- [ ] `shell/platform/android/android_exports.lst` — drop the ghost entries `shorebird_active_path` and `shorebird_active_patch_number` (no longer exist in Rust) and `shorebird_check_for_update` (removed in this PR).

Until the engine-side PR lands, an engine build pulling this updater commit will fail to find `updater.h`. Please coordinate the merge order.

## Test plan

- [x] `cargo test` — 227/227 pass
- [x] `cargo clippy --all-targets` — clean
- [x] `cargo fmt --check` — clean
- [x] `flutter test` in `shorebird_code_push` — 40/40 pass
- [x] `dart analyze lib test` — clean (one pre-existing line-length info, unrelated)
- [x] Verified all 9 symbols `flutter::shorebird::RealUpdater` calls are present in `updater_engine.h`
- [x] Verified ffigen-generated bindings now expose only the 5 Dart-stable symbols
- [x] Both new headers parse cleanly as C and C++


eseidel:
I was looking to see what it would take for us to make the engine restartable, Claude was getting confused about what was stable public api (dart) vs. not (engine) so I asked it to split the two.

eseidel:
There should be no functional changes from this other than removing some dead code. This will require a one-line update to Flutter Engine when the next role happens.

library/src/c_api/dart.rs
```diff
@@ -0,0 +1,138 @@
...
+pub extern "C" fn shorebird_check_for_downloadable_update(c_channel: *const c_char) -> bool {
+    log_on_error(
+        || {
+            let channel = to_rust_option(c_channel)?;
+            updater::check_for_downloadable_update(channel.as_deref())
+        },
+        "checking for update",
+        false,
+    )
+}
```
bdero, line 106: Not a blocker for this PR, but are we exercising all of the possible shorebird code push FFI bindings in CI today? My guess would be no, and I should probably figure out how to get this integration tested (perhaps in the e2e smoketests if we can't figure out something lighter) so that it's not possible for us to break them by accident in a prod release.

bdero (approved):

library/src/c_api/dart.rs
eseidel: I talked with claud about it, specifically about setup integration tests at the Shorebird_code_push layer. It's definitely doable. I think probably ideally we end up testing them in an app layer with real desktop releases.

eseidel: In short, no we are not yet, but could.

PR #352, opened 2026-05-05
refactor: per-patch lifecycle state machine

### LINKED ISSUE #3737: tech debt: consolidate per-patch state machine in updater
## Motivation

The updater tracks the lifecycle of each patch number across at least six separate stores:

- `{patch_number}.download.json` sidecar (URL, expected hash, expected size)
- The raw `{patch_number}` download file (existence + size)
- The inflated `{patch_number}.full` file (existence)
- `is_known_bad_patch` list in `UpdaterState` (boot failures only today)
- `next_boot_patch` / `current_boot_patch` in `UpdaterState`
- The events queue

There is no single source of truth for "what state is patch N in?" — we infer it from cross-referencing these stores. New behaviors get added by introducing new orthogonal axes (e.g. \"is sidecar size > file size?\", \"is patch in known-bad list AND has partial download?\"), which multiply the state space instead of moving the patch through defined transitions.

## Concrete trigger

While debugging a customer 416 loop (shorebirdtech/updater#351) we identified a follow-on bug: when `check_hash` fails after install, the bytes are deterministically bad on this device (typically a base-binary mismatch on the customer's side), but we currently re-download on every update cycle. The natural fix — mark the patch bad on install hash mismatch — was deliberately deferred because adding it on top of today's scattered state would create yet another orthogonal interaction (\"known-bad AND has-partial-file\", \"known-bad AND has-completed-download\", etc).

## Proposed direction

Model an explicit per-patch state machine. Rough states:

- `Unknown` (no record)
- `Downloading { resume_from }` (sidecar + partial file)
- `Downloaded` (sidecar + file size matches expected_size, not yet inflated)
- `Inflated` (`.full` file present, hash-checked)
- `Installed` (in `next_boot_patch`)
- `Booted` (in `current_boot_patch`)
- `Bad { reason }` (boot failure, install hash mismatch, etc.)

Persist one document per patch number capturing this state, instead of inferring it from cross-cutting file-existence checks. Transitions become explicit functions; behaviors like \"don't re-download a known-bad patch\" fall out naturally.

## Behaviors that fall out

- Mark patch bad on install hash mismatch → no redownload loop (the customer-impacting follow-on from #351).
- Mark patch bad after N consecutive transient failures (configurable threshold).
- Garbage collect orphan files based on declared state, not heuristics.
- Cleaner `should_install_patch` — single state-check, no scattered preconditions.

## Out of scope

The compatibility/migration story for existing on-disk state is non-trivial — several million devices have current-format sidecars and `UpdaterState` files. Migration plan is part of the design, not glossed over.

## Related

- shorebirdtech/updater#351 (the 416 fix, scoped to avoid expanding into this)

eseidel:
Closes shorebirdtech/shorebird#3737.

## Why

Per-patch state was scattered across six stores — the `download_state.rs` sidecar JSON, the bare files in `downloads/`, the inflated `.full` file, the `patches/{N}/dlc.vmcode` artifact, the `next_boot_patch` / `last_booted_patch` / `currently_booting_patch` / `known_bad_patches` fields of `PatchesState`, and the session-only `running_patch` global. There was no single source of truth for "what state is patch N in?"; every behavior change (most recently the 416 fix in #351) added another orthogonal axis instead of moving the patch through defined transitions.

This PR consolidates the persistent per-patch state into a single document at `{cache}/patches/{N}/state.json` driven by an explicit state machine.

## On-disk layout (per release)

```
{cache}/
  state.json                    # {client_id, release_version, queued_events}
  pointers.json                 # ReleasePointers (patch numbers only)
  patches/
    {N}/
      state.json                # PatchState — source of truth for patch N
      download                  # compressed bytes (Downloading/Downloaded only)
      dlc.vmcode                # installed artifact (Installed only)
```

Everything under `patches/` is wiped on release-version change. Within a release, `state.json` survives as a `Bad` tombstone even after artifact files are removed — that's how rollback to a previously-known-bad patch is prevented when newer patches fail.

## State machine

```rust
enum PatchState {
    Downloading { url, hash, signature, partial_size },
    Downloaded  { url, hash, signature, size },
    Installed   { hash, signature, size },
    Bad         { reason, hash?, signature?, size? },   // tombstone
}

enum BadReason {
    BootCrash,            // boot started but never recorded success
    InvalidPatchBytes,    // inflate failed (zstd magic / decompression)
    InstallHashMismatch,  // bytes decompressed but produced the wrong hash
    ValidationFailed,     // size or signature check failed at boot time
}
```

`pointers.json` carries only patch numbers (`next_boot_patch`, `last_booted_patch`, `currently_booting_patch`, `boot_started_at`). Metadata lives once per patch in its `state.json`.

## API

The lifecycle exposes two state-aware operations callers don't have to choose between:

- **`mark_bad(n, reason)`** — writes a `Bad` tombstone and removes artifact files. Sugar over `write_state(Bad{...}) + cleanup`.
- **`cleanup(n)`** — state-aware retirement. If currently `Bad`, keeps the tombstone and removes only artifact files. Otherwise removes the entire patch directory.

Plus the lifecycle transitions used by `update_internal` and `report_launch_*`:

- `decide_start(n, url, hash) → DownloadAction { Fresh, Resume{offset}, Complete, Skip }`
- `record_download_started`, `record_download_complete`, `record_install_complete`
- `record_boot_start`, `record_boot_success`, `record_boot_failure`
- `detect_boot_crash_on_init` (runs on init; promotes a stranded `currently_booting_patch` to `Bad{BootCrash}`)
- `validate_next_boot_patch` (size + signature; marks `Bad{ValidationFailed}` on failure)
- `recompute_next_boot` (state-aware fallback walk that doesn't clobber a valid newer `next_boot_patch`)

## Behaviors that fall out

- **416 redownload loop** (#351): `decide_start` returning `Complete` skips the network when bytes are already on disk; `mark_bad(InstallHashMismatch)` short-circuits subsequent cycles via `Skip(KnownBad)`. The post-#351 redownload loop (every cycle: download → hash mismatch → cleanup → repeat) is now bounded to one cycle per `(release, patch_number)`.
- **Server rollback un-rollback**: `uninstall_patch` calls `cleanup` (state-aware), not the old "always-add-to-known-bad" path. A patch the server later un-rolls-back will be redownloaded — but a patch that crashed locally and is then rolled back keeps its `Bad{BootCrash}` tombstone, so the server's "it's fine again" can't override device-local "this crashed."
- **Chunked-encoding resume**: `decide_start` reads the on-disk file size as the resume offset, not a recorded count, so chunked-encoded responses (no `Content-Length`) work the same as fixed-length ones. The state.json is written before the download starts.
- **OS-evicted partial files**: `decide_start` cross-checks the download file's existence for `Downloading` / `Downloaded` states. If the file was evicted, the state is unrecoverable — fall through to `Fresh`.

## Diff scope

```
library/src/cache/lifecycle.rs     | 1365 +++  (new)
library/src/cache/patch_manager.rs | 1615 ---  (deleted)
library/src/download_state.rs      |  126 ---  (deleted)
library/src/cache/updater_state.rs |  622 +-   (rewritten on top of lifecycle)
library/src/updater.rs             |  914 +-   (update_internal cutover, helpers removed)
library/src/cache/mod.rs           |    4 +-
library/src/lib.rs                 |    1 -
7 files changed, 1890 insertions(+), 2757 deletions(-)
```

Net `−867` lines of code despite adding a full state-machine module — the patch_manager / download_state / scattered-state-helpers tonnage outweighs it.

## Test plan

- [x] `cargo test` — 213 tests pass (down from 257 because the patch_manager unit tests are gone with the module). End-to-end coverage in `updater.rs::tests` is unchanged and all green.
- [x] 67 new lifecycle unit tests covering: serde roundtrip, corrupt state.json fail-safe, mark_bad metadata preservation across all source states, cleanup state-awareness (Bad → tombstone, else → forget), idempotency, decide_start across all states + URL/hash mismatches + missing-file recovery, all download/install transitions, full boot lifecycle (start/success/failure), crash-on-init detection, fallback walk, validation.
- [x] Behavior of every `updater::tests` resume / rollback / state-recovery / install-failure scenario reproduced end-to-end against the new layout.
- [x] `cargo fmt --check` clean.

## Caveats

- **Bumps the on-disk format.** Per the design discussion, everything under `{cache}/patches/` is wiped on release-version change — no migration needed because per-release state can change freely between releases. Devices upgrading to a release shipped with this PR will redownload any in-flight patch once. This is consistent with how per-release state was always handled.
- **`config.download_dir` is now vestigial.** The lifecycle uses `{storage_dir}/patches/{N}/` so the prior `code_cache_dir/downloads` location is no longer used. The field is left in `UpdateConfig` to keep the C API stable; it can be removed in a follow-up.
- **`FileOperation::DeleteDir` is now only used by tests.** Compile-time warning; safe to ignore or remove the variant in a follow-up.


eseidel:
@bdero asked for us to stand up end-to-end testing first.

eseidel:
Built an integration test system per your request: https://github.com/shorebirdtech/updater/pull/353

library/src/cache/lifecycle.rs
```diff
@@ -0,0 +1,2165 @@
...
+    /// `currently_booting_patch`) to match the prior PatchManager API
+    /// shape — most call sites already have the number in hand. The
+    /// breadcrumb is cleared regardless of whether it matched.
+    pub fn record_boot_failure(&mut self, n: usize) -> Result<()> {
+        self.pointers.currently_booting_patch = None;
+        self.pointers.boot_started_at = None;
+        self.save_pointers()?;
+        self.mark_bad(n, BadReason::BootCrash)?;
+        self.recompute_next_boot()
+    }
```
bdero, line 581: _bdero's review bot_: 💡 clearing the `currently_booting_patch` breadcrumb *before* `mark_bad` opens a small crash window: if the process dies between `save_pointers` and `mark_bad`, the patch stays `Installed` and `detect_boot_crash_on_init` won't fire next boot because the breadcrumb is gone. Worth flipping the order — `mark_bad` first, then clear the breadcrumb — so the worst case is "next init detects a still-set breadcrumb and re-marks Bad" (idempotent) instead of "patch silently gets retried."

library/src/cache/lifecycle.rs
```diff
@@ -0,0 +1,2165 @@
...
+            if self.pointers.next_boot_patch != new_target {
+                self.pointers.next_boot_patch = new_target;
+                dirty = true;
+            }
+        }
+        if dirty {
+            self.save_pointers()?;
+        }
+        Ok(())
+    }
```
bdero, line 670: _bdero's review bot_: 💡 `recompute_next_boot` only considers `last_booted_patch` as a fallback target. The doc comment calls this out as deliberate ("at most a couple of patches active at once") — fair for the common case, but if a `Bad{ValidationFailed}` next_boot has an Installed predecessor *that was never booted* (e.g. user upgraded build, the freshly Installed patch fails validation, and there's nothing in `last_booted_patch` because this is a fresh release), we go straight to base release even though there might be an installable older patch sitting in the dir. Probably acceptable; flagging since the doc could be more explicit that this is "fall back to last_booted only" rather than "fall back to anything bootable."

library/src/cache/lifecycle.rs
```diff
@@ -0,0 +1,2165 @@
...
+            n,
+            &PatchState::Bad {
+                reason,
+                hash,
+                signature,
+                size,
+            },
+        )?;
+        self.cleanup(n)
+    }
```
bdero, line 273: _bdero's review bot_: 💡 `mark_bad` reads state, writes Bad, then calls `cleanup` which reads state again. Two disk reads per mark_bad transition. Not a perf concern at scale, but `cleanup` could take a `state: &PatchState` to skip the second read. Or `mark_bad` could call `delete_artifact_files` directly since it just wrote `Bad` and knows the post-state.

library/src/cache/lifecycle.rs
```diff
@@ -0,0 +1,2165 @@
...
+            if let Some(public_key) = public_key {
+                let signature = signature.context("Patch signature is missing")?;
+                let actual_hash = signing::hash_file(&path)?;
+                signing::check_signature(&actual_hash, &signature, public_key)?;
+            } else {
+                shorebird_info!("No public key configured; skipping signature verification");
+            }
+        }
+        Ok(())
+    }
```
bdero, line 725: _bdero's review bot_: 💡 `validate_installed_patch` checks size + (in Strict) signature, but the `Installed.hash` field never gets compared against the on-disk artifact's hash. Worth documenting that explicitly — the hash check happens at install time and isn't re-run at boot — since a reader staring at the recorded `hash` in the struct will reasonably wonder why it's there if not for this. Or fold it into the validation; recomputing sha256 on a 100KB-1MB artifact is fast.

library/src/cache/updater_state.rs
```diff
@@ -1,63 +1,80 @@
...
+/// Files and directories under `cache_dir` that shorebird has ever
+/// written. On a release-version change or unparseable state we wipe
+/// these and keep going. `state.json` is intentionally absent — it's
+/// rewritten in place with the preserved `client_id` (so removing it
+/// would lose the only thing we want to carry forward).
+///
+/// `patches_state.json` is the legacy file from the prior `PatchManager`
+/// implementation; carrying it forward would orphan ~few-KB of stale
+/// state on every device upgrading through this PR.
+const SHOREBIRD_OWNED_PATHS: &[&str] = &["patches", "pointers.json", "patches_state.json"];
```
bdero, line 34: _bdero's review bot_: 💡 `SHOREBIRD_OWNED_PATHS` works but relies on devs remembering to update it when adding files under `cache_dir`. The doc on `create_new_and_save` flags this, which helps. A more enforce-y option: a single `paths()` method on a "files-shorebird-writes" type that's the only way to write under cache_dir. Probably not worth the churn for a 3-entry list, but worth keeping in mind as the count grows.

library/src/cache/lifecycle.rs
```diff
@@ -0,0 +1,2165 @@
+//! Per-patch lifecycle state machine.
```
bdero, line 1: _bdero's review bot_: 🧹 2165 lines in one file. Splitting into `lifecycle/state.rs` (types), `lifecycle/transitions.rs` (record_*), `lifecycle/cleanup.rs` (cleanup_*) would help navigation. Not blocking — premature splitting is also bad — but a candidate for a follow-up once the surface stabilizes.

library/src/cache/lifecycle.rs
```diff
@@ -0,0 +1,2165 @@
...
+            std::fs::remove_dir_all(&dir)?;
+        }
+        let download = self.download_artifact_path(n);
+        if download.exists() {
+            if let Err(e) = std::fs::remove_file(&download) {
+                shorebird_error!("Failed to remove {:?}: {:?}", download, e);
+            }
+        }
+        Ok(())
+    }
```
bdero, line 332: _bdero's review bot_: 🧹 `delete_artifact_files` and `forget_dir` both have the same compressed-download-removal stanza. Could factor into a `remove_download_file(n)` helper.

library/src/cache/updater_state.rs
bdero, line 34: _bdero's review bot_: 🧹 `patches_state.json` in the wipe list is good migration hygiene, but worth a `TODO` with a target version for removal so we don't carry it forever.

library/src/cache/lifecycle.rs
```diff
@@ -0,0 +1,2165 @@
...
+//!
+//! Mutations are exposed as two operations on top of the raw read/write:
+//!   - `mark_bad(n, reason)` writes a Bad tombstone and deletes artifact
+//!     files (sugar over `write_state` + `cleanup`).
+//!   - `cleanup(n)` is state-aware: keeps the tombstone if the patch is
+//!     already Bad, otherwise removes the patch directory entirely.
+//!
+//! Callers never pick between "delete tombstone" and "preserve tombstone";
+//! the state on disk decides. See the design notes that led here in
+//! shorebirdtech/shorebird#3737.
```
bdero, line 37: _bdero's review bot_: 🧹 module docstring describes pointers + per-patch state but doesn't mention that `running_patch` lives separately in `config.rs` because it's session-scoped. Worth a sentence so a reader looking for "where does `shorebird_current_boot_patch_number` get its value" doesn't go grepping inside lifecycle.rs.

library/src/cache/lifecycle.rs
```diff
@@ -0,0 +1,2165 @@
...
+    pub last_booted_patch: Option<usize>,
+
+    /// Boot-in-progress breadcrumb. Set at `record_boot_start`, cleared
+    /// at `record_boot_success` / `record_boot_failure`. If still set on
+    /// next init, treat as a crashed boot.
+    pub currently_booting_patch: Option<usize>,
+
+    /// Unix timestamp (seconds) when `currently_booting_patch` was set.
+    pub boot_started_at: Option<u64>,
+}
```
bdero, line 145: _bdero's review bot_: 🧹 naming: `currently_booting_patch` (boot-in-progress breadcrumb) and `running_patch` (session-scoped FFI signal) are *different* concepts but the names are close enough that a reader has to lean on the comments to keep them apart. Not introduced by this PR, but the cleanup commits could rename one of them — `boot_in_progress_patch` for the breadcrumb would make the distinction obvious.

bdero (commented):
## Overall

Right architecture, right time. Consolidating the six scattered state stores into one per-patch state.json driven by an explicit machine is the move, and `cleanup` being state-aware (Bad → keep tombstone, else → forget entirely) cleanly separates the two prior conflated concerns ("server told us not to use this" vs "this actually crashed on this device"). #356 falls out as a strict subset once this lands — happy to close it.

A handful of suggestions, no blockers. Inline comments tagged 💡 (suggestion) and 🧹 (nit).

## 💡 No end-to-end test for the customer scenario

`library/src/updater.rs` doesn't have an end-to-end test that pulls the customer scenario in shorebirdtech/shorebird#3728 through `update_internal` (install → server `rolled_back: [1]` → server returns patch 1 again with `rolled_back: []` → assert reinstall). The behavior is correct at the lifecycle unit level (`cleanup_on_non_bad_patch_forgets_entirely`), but the customer  reported it through the FFI, so a test at that level would directly nail the regression we just shipped a hotfix for in #356. Cheap to port from #356.

(Inlining this in the body since `updater.rs` line ~2229 — the natural anchor next to the existing rollback tests — isn't in a diff hunk.)

bdero (approved):
passes onceover for me, bot found some minor suggestions. Let's merge today and I'll launch a prod smoke test for it.

PR #353, opened 2026-05-05
feat: stage 1 integration test harness

eseidel:
## Summary

Stage 1 of the integration test harness described in `docs/integration_tests.md` (also in this PR). Drives the Dart `Updater` API through real FFI into a real cdylib that re-exports the production C API alongside test-only hooks. One trivial test today; stages 2 (FakePatchServer + golden path) and 3 (adversarial scenarios) land separately.

## What's in here

**New `library_test_hooks` cdylib crate**
- Depends on `updater` with a new `test-hooks` Cargo feature that widens internal Rust `pub` items (today: `testing_reset_config`) for sibling-crate access only. Production builds — the cdylib/staticlib that ships in the engine — do not enable the feature, and `nm` confirms `shorebird_test_reset` does not appear in the production cdylib.
- Re-exports `updater::c_api::dart::*` and `updater::c_api::engine::*` so the rlib's `#[no_mangle]` symbols flow into the resulting cdylib (no DCE intervention needed). One artifact, both surfaces.
- Excluded from workspace `default-members` so plain `cargo build` / `cargo test` doesn't unify the feature into production builds.

**Dart-side harness**
- `test/integration/all_test.dart` — single file (header comment explains the same-isolate / `OnceCell` constraint). `setUpAll` shells out to `cargo build -p library_test_hooks`, opens the cdylib with `DynamicLibrary.open`, and reassigns the existing `@visibleForTesting` `Updater.bindings`. If cargo isn't available, `markTestSkipped` keeps the suite green.
- `ffigen_test_hooks.yaml` generates `TestHooksBindings` under `test/integration/generated/`. Test-only bindings stay out of `lib/`.
- Zero changes to `shorebird_code_push/lib/`.

**CI**
- `library_test_hooks` is added to the rust_crate matrix (Linux, macOS, Windows).
- `shorebird_code_push` build now triggers on `library/**` and `library_test_hooks/**` too, so the integration suite catches cdylib regressions.

**Doc**
- `docs/integration_tests.md` — the design write-up. Updated through review to reflect the test-hooks crate factoring, the existing `Updater.bindings` test seam, single-file process model, and runtime skip via `markTestSkipped`.

## Test plan

- [x] `cargo test -p updater` — 232 passed locally.
- [x] `dart test` in `shorebird_code_push/` — 42 passed (41 existing unit tests + 1 new integration test).
- [x] `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, `dart format`, `dart analyze --fatal-warnings lib test`, cspell — all clean.
- [x] `nm target/debug/libupdater.dylib | grep shorebird_test_reset` returns nothing (production cdylib does not contain test symbols).
- [ ] CI green on Linux + macOS + Windows for the rust_crate matrix.
- [ ] CI green for the shorebird_code_push integration test (ubuntu runners have Rust pre-installed; if not, the test skips rather than fails).


bdero (approved):
lgtm

PR #356, opened 2026-05-07
fix: don't permanently ban patch numbers on server-driven rollback

bdero:
## Summary

Customer report: after rolling a patch back and then rolling it forward from the dashboard, the device silently refuses to install the re-offered patch. `checkForUpdate` returns `upToDate` while the server still says a patch is available, and FRC-driven notification logic that compares `currentPatch` against a target patch number loops on a force-restart prompt that nothing can satisfy.

## The bug

PR #348's review-feedback commit ([59bb984](https://github.com/shorebirdtech/updater/commit/59bb9843f21b02b9d69776f56dc7b614981a2abd)) added `known_bad_patches.insert(patch_number)` to `remove_patch` to defend against an orthogonal resurrection-via-failed-delete race. That set is a permanent ban for the current `release_version`, so the same number also stays banned through:

- **A dashboard rollforward.** `rollforward.dart:38-42` flips `is_rolled_back` back to `false` on the same `app_release_patch_` row, so the server resends exactly the same `Patch` (same number, same hash).
- **A delete-and-recreate of the release at the same `release_version`.** This is the workflow `shorebird_cli/lib/src/commands/release/release_command.dart:518` itself prescribes when bumping Flutter. New patches start at #1 again with new hashes but inherit the old release's bans on those numbers.

The customer log line `[shorebird] Patch 1 has previously failed to boot, skipping` is misleading — the patch never boot-failed, the rollback signal poisoned the number.

## The fix

In `remove_patch`, drop `last_booted_patch` when its number matches the rolled-back patch instead of inserting into `known_bad_patches`. That keeps the resurrection defense — the else-if fallback in `try_fall_back_from_patch` has no target to promote — and also stops banning the number for any later legitimate reuse.

Safe to clear `last_booted_patch` here because `running_patch` (added in #348) is session-scoped and is now the FFI's source of truth for "what's running this process," so dropping the persisted historical pointer doesn't regress the patch-to-release rollback fix.

| Field | Role after this PR |
|---|---|
| `last_booted_patch` | Fallback target for `try_fall_back_from_patch`. Dropped on server rollback of the same number. |
| `next_boot_patch` | What the engine should load next launch. `None` = base release. |
| `running_patch` | Session-scoped: what this process is using. FFI source of truth. Unchanged. |
| `known_bad_patches` | Patches that *actually failed to boot*. No longer used for server-driven rollback. |

## Verification

- **New regression test** `rollforward_after_rollback_can_install_same_patch` in `library/src/c_api/mod.rs`. Reproduces the customer scenario at the C API level: install patch 1, server rolls patch 1 back, server rolls patch 1 forward (same hash, empty `rolled_back`). Pre-fix: `check_for_downloadable_update` returns false and the log line matches the customer's logs. Post-fix: returns true and the patch is queued for next boot.
- **Updated** `rollback_then_failed_replacement_does_not_resurrect_rolled_back_patch` to assert the property the test name promises (patch 1 is not promoted to `next_boot_patch`) plus the new invariant (`is_known_bad_patch(1)` is false, so a rollforward can install it). The original assertion was checking the implementation mechanism rather than the behavioral property.
- All 235 existing tests pass without modification. `cargo clippy --lib --tests` clean. `cargo fmt --check` clean.

## Out of scope (follow-up)

This PR doesn't fix the broader weakness that `known_bad_patches` is keyed by `patch_number` alone — a number that isn't globally unique (it can refer to entirely different artifacts under the same `release_version` if the release is deleted-and-recreated). The wire response already carries `Patch.hash` and the server-side patch id, so a follow-up can key the bad-patch set by hash and make this whole class of bug impossible by construction.

## Test plan

- [x] Regression test fails on `origin/main`, passes on this branch
- [x] All existing tests pass without modification
- [x] `cargo test --lib` 235 pass / 0 fail
- [x] `cargo clippy --lib --tests` clean
- [x] `cargo fmt --check` clean
- [ ] Validate on a real device against the customer scenario before landing

Refs: shorebirdtech/shorebird#3728, #348

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -67,8 +67,10 @@ words:
   - pubspec
   - repr
   - reqwest
+  - resends
   - rlib
   - rsplit
+  - rollforward
   - rollouts
   - RTLD
   - rustflags
--- a/library/src/c_api/mod.rs
+++ b/library/src/c_api/mod.rs
@@ -876,6 +876,114 @@ mod test {
         assert_eq!(shorebird_current_boot_patch_number(), 0);
     }
 
+    /// Regression: rolling a patch forward after the server rolled it back
+    /// must allow the device to install it again.
+    ///
+    /// The dashboard's rollback and rollforward operations toggle
+    /// `is_rolled_back` on the same `app_release_patch_` row, so the server
+    /// resends the same `Patch` (same number, same hash) on rollforward.
+    /// `remove_patch` previously inserted the patch number into
+    /// `known_bad_patches` to defend against an orthogonal
+    /// resurrection-via-failed-delete race; that insert also banned the
+    /// number for any later legitimate reuse, so `should_install_patch`
+    /// returned `PatchKnownBad` for the rolled-forward patch and the
+    /// device silently refused it.
+    ///
+    /// Customer-visible symptom: `checkForUpdate` returns `upToDate` while
+    /// the server still says a patch is available. Combined with FRC-driven
+    /// notification logic that compares `currentPatch` against a target
+    /// patch number, the app loops on a force-restart prompt that nothing
+    /// can satisfy.
+    #[serial]
+    #[test]
+    fn rollforward_after_rollback_can_install_same_patch() {
+        testing_reset_config();
+        let tmp_dir = TempDir::new().unwrap();
+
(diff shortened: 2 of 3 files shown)


eseidel:
I believe this is also fixed by https://github.com/shorebirdtech/updater/pull/352.

PR #357, opened 2026-05-10
feat(updater): Draft future-release patch prefetch

### LINKED ISSUE #3755: feat: prefetch patches for server-advertised future releases
## Problem

When a device auto-updates the native binary to a newer release that already has patches published, those patches aren't on disk yet. On first launch of the new binary, the user runs unpatched code even though a fix exists. This happens because [`/api/v1/patches/check`](https://github.com/shorebirdtech/updater/blob/main/library/src/network.rs#L16) is scoped to a single [`release_version`](https://github.com/shorebirdtech/updater/blob/main/library/src/network.rs#L242); patches for any other release on the same track aren't fetched until the user is already running that release.

<p align="center">
  Image
</p>

Concrete scenario:

Consider an app that has version 9.0.0 in production, and *9.0.1 is pending App Store review*.

- A user reports a data-loss bug that is affecting the production release, 9.0.0.
- Dev releases a hotfix for the bug, as patch 9.0.0-p1. Since 9.0.1 is already pending review, rather than repackaging the release and resubmitting, they patch that release as well, with 9.0.1-p1.
- Dev tells user to restart to apply patch 9.0.0-p1; user does so and is pleased with the fix. User goes to bed.
- Meanwhile, v9.0.1 is approved by App Review and gets installed as an automatic update for the user overnight.
- User boots v9.0.1 in the morning. Since it's the first launch of that release, and the shorebird client didn't pre-download the v9.0.1-p1 patch ahead of time, it's not available to be applied on that first launch of 9.0.1. The update checker downloads it in the background, and it's ready to apply (pending restart), but the user doesn't know that.
- As a result, the user encounters the "fixed bug" again.
  - User says: "My data is gone again; I thought you said you fixed it? what do you mean the patch 'un-applied?'", etc.
  - Result: Broken trust and confusion (justifiably)

## Proposal

Pre-fetch patches for server-advertised release versions on the same track, so when the OS swaps in a new native binary, the patch can apply on first launch instead of the user booting unpatched code.

<p align="center">
  Image
</p>

Nuances:
  - Benefits existing users on every native update, in addition to the emergency-hotfix scenario.
  - Doesn't depend on solving the harder fresh-install-boots-unpatched problem ([#741 ("feat: install patches without having to restart the app")](https://github.com/shorebirdtech/shorebird/issues/741), consolidated into [#2350 ("feat: enable advanced users to implement patch-install-without-restart")](https://github.com/shorebirdtech/shorebird/issues/2350)).
- This isn't updater-only, but the updater already has most of the local primitives this needs: the engine asks the updater which patch to boot, patch state is release-scoped, and boot validation happens against the running binary.
  - The missing pieces appear to be: server/API support to advertise server-selected prefetch candidates for the same app/track/platform/architecture, public protocol support for that response field, updater storage that survives the release-version reset, and updater promotion logic that validates the prefetched delta after the matching native binary is running.
  - The engine ([`shorebird_next_boot_patch_path()`](https://github.com/shorebirdtech/engine/blob/shorebird/dev/shell/common/shorebird/shorebird.cc#L82-L86)) consumes whatever path the updater designates, so a patch designated during initialization can be picked up on launch.
  - [Strict-mode boot validation re-verifies the signature at launch](https://github.com/shorebirdtech/updater/blob/main/library/src/cache/lifecycle.rs#L80-L82) using the running binary's public key. In this proposal, the current release stores only the compressed delta for a server-advertised future release, and the future release promotes/validates it only after its native binary is actually running.
  - The disk schema is already release-scoped, but prefetched compressed deltas need a location that survives the release-version reset.
  - One concrete blocker is the [unconditional wipe-on-mismatch](https://github.com/shorebirdtech/updater/blob/main/library/src/cache/updater_state.rs#L187-L207) in `load_or_new_on_error()`, which deletes patches when `release_version` changes. The updater draft PR preserves a dedicated prefetch area across that reset and promotes matching bytes on first init of the new release.
- Default-on, with an opt-out flag for unusual cases (e.g. `prefetchUpcomingPatches: false` for metered networks or embedded devices). The cost is small and bounded; the benefit applies to every native update for every user; opt-in adoption would leave most apps vulnerable to the regression by accident.
  - Per-patch bandwidth is small: Android tens of KB to ~166 KB ([#2737 ("feat: make patch sizes smaller on Android")](https://github.com/shorebirdtech/shorebird/issues/2737)), iOS hundreds of KB to a few MB ([#811 ("feat: Make patch sizes smaller on iOS")](https://github.com/shorebirdtech/shorebird/issues/811)).
  - The simplest cache policy bounds disk usage to ~2 patches at any time: keep the patch for the current release plus the patch for the latest server-advertised release on the track.
    - Self-limits across release churn: a user on `9.5.0-p1` with `9.5.1-p1` pre-fetched will see `9.5.1-p1` quietly replaced by `9.5.2-p1` when the dev ships 9.5.2.
    - Even a user squatting on an old native binary indefinitely costs no more than one extra patch's worth of disk.
  - The regression we're avoiding is invisible to the user; they just see a previously-fixed bug come back.

## Draft PRs

- [shorebirdtech/updater#357](https://github.com/shorebirdtech/updater/pull/357): client-side draft/proof-of-concept for preserving prefetched patch bytes across release reset and promoting them when the matching release first runs.
- [shorebirdtech/shorebird#3781](https://github.com/shorebirdtech/shorebird/pull/3781): public protocol sketch for advertising prefetch candidate release versions. This is closed for now pending discussion here, but included to show the shape of the protocol piece.

Both drafts assume private server/API support to populate the prefetch candidates, or an equivalent server-owned mechanism.

## Related

- [#1980 (open, "feat: expose a way to detect newer releases")](https://github.com/shorebirdtech/shorebird/issues/1980): Complementary, dev-side; with this proposal you wouldn't need to nag users to update just to avoid this patch-continuity gap.
- [#2350 (open, "feat: enable advanced users to implement patch-install-without-restart")](https://github.com/shorebirdtech/shorebird/issues/2350): Complementary at the in-session level. This proposal addresses what happens *across native binary swaps*.
- [#3702 (merged, "feat(protocol): add `current_patch_number` to PatchCheckRequest")](https://github.com/shorebirdtech/shorebird/pull/3702): Shows the protocol is already evolving in this area.
- [#3749 (open, "feat: ShorebirdUpdater.watchForUpdates, managed polling for manual-update apps")](https://github.com/shorebirdtech/shorebird/issues/3749): Today scoped to the current release; this proposal would naturally extend it.

lukemmtt:
Part of a proposed solution to close shorebirdtech/shorebird#3755, along with  shorebirdtech/shorebird#3781.

Note: This remains a draft client-side proof of concept for the updater portion. It becomes merge-candidate only after the Shorebird API/server can advertise prefetch candidates, and after the public protocol surface is regenerated from that server/OpenAPI source of truth.

## Summary

When a user's native binary auto-updates between Shorebird patch checks, the device today must boot the unpatched new release until the next patch check completes. This change lets the running binary park raw deltas for server-advertised sibling `release_version`s on the same track, so that when the user later updates to one of them, the matching delta is already on disk and can be promoted into the active boot path during init.

This is not intended to solve fresh installs booting unpatched, or in-session patch activation. It is the narrower existing-user continuity gap across native store updates.

## Approach

Built on top of the per-patch lifecycle state machine introduced in #352. A prefetched patch is stored as a `PatchState::Downloaded` document in a parallel namespace keyed by `release_version`:

```text
<storage_dir>/prefetched_patches/<rv>/patches/<n>/state.json
<storage_dir>/prefetched_patches/<rv>/downloads/<n>
```

The compressed bytes intentionally live under `storage_dir`, not `download_dir`, because release-version resets wipe the active download root before promotion runs for the newly updated binary. This fixes the main correctness issue in the earlier draft.

The prefetched namespace is independent of the per-release `patches/` and `download_dir/<n>` paths the active lifecycle owns. It still uses a `PatchLifecycle`, which lets the prefetch path reuse `record_download_started`, `record_download_complete`, `read_state`, and other downloaded-state transitions without sharing the active release's patch directory.

## Wire shape

- Server/API side: extend `PatchCheckResponse` with optional `available_release_versions: Vec<String>`, or an equivalent server-owned prefetch-candidate field, containing release versions on the requesting app's same app/channel/platform/arch track that are useful and safe to prefetch. Older clients ignore the field.
- Public protocol side: shorebirdtech/shorebird#3781 adds the corresponding Dart protocol surface as `PatchCheckResponse.availableReleaseVersions`. That PR is also draft because this package is generated from the public OpenAPI schema, which does not expose the field yet.
- Client side: after the existing patch-check and install flow runs for the current release, iterate over the advertised siblings, excluding the current `release_version`, issue one extra `/patches/check` per sibling with `release_version` overridden, download the raw delta, and record it as `PatchState::Downloaded` in the prefetched namespace.
- Promotion: `handle_prior_boot_failure_if_necessary` also tries to promote a prefetched delta whose release version matches the now-running binary. The compressed bytes are moved into the active download root, a `Downloaded` state is written into the active `patches/<n>/`, then the existing `inflate -> check_hash -> record_install_complete -> promote_to_next_boot` pipeline runs against it.

No inflation or hash/signature verification happens at prefetch time, because the running binary is the wrong inflation base for any release other than its own. The new binary's `patch_public_key` is the right key for hash verification and Strict-mode boot signing at promotion/boot time.

## Lifecycle and cleanup

- Cached siblings the server stops advertising are pruned via `retain_prefetched_release_versions` on every successful patch check.
- Prefetched bytes from the earlier draft path under the active download root are also cleaned up when encountered.
- Missing or evicted prefetched bytes fail safely and fall back to the normal patch-check flow.
- Prefetch errors are best-effort and never affect the main update result.

## Open items for the Shorebird team

- Private API/server change to populate `available_release_versions`, or an equivalent field, from server-selected same-track release candidates.
- Public OpenAPI/protocol regeneration so shorebirdtech/shorebird#3781 comes from the source of truth instead of a manual generated-file edit.
- Decide whether sibling patch checks should include an explicit prefetch marker to avoid distorting release activity analytics.
- Decide whether candidate selection should respect rollout bucketing up front. This branch still asks the server again per sibling, so per-release rollout decisions can also happen during the sibling check.

## Test plan

- `cargo fmt --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test --workspace`

Local result: `cargo test --workspace` passes with 250 updater tests plus the patch crate test.

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -66,6 +66,7 @@ words:
   - msvc
   - nscloud
   - oslog
+  - patchless
   - pubspec
   - repr
   - reqwest
--- a/library/src/c_api/mod.rs
+++ b/library/src/c_api/mod.rs
@@ -428,6 +428,7 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: None,
+                    available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_TESTS_PATCH.write_to(dest),
@@ -482,6 +483,7 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: None,
+                    available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_TESTS_PATCH.write_to(dest),
@@ -539,6 +541,7 @@ mod test {
                     patch_available: false,
                     patch: None,
                     rolled_back_patch_numbers: None,
+                    available_release_versions: None,
                 })
             },
             UNEXPECTED_DOWNLOAD,
@@ -623,6 +626,7 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: None,
+                    available_release_versions: None,
                 })
             },
             |_url, _dest: &Path, _resume_from: u64| Err(anyhow::anyhow!("Error")),
@@ -673,6 +677,7 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: None,
+                    available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_TESTS_PATCH.write_to(dest),
@@ -739,6 +744,7 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: None,
+            available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_TESTS_PATCH.write_to(dest),
@@ -768,6 +774,8 @@ mod test {
                     patch_available: false,
                     patch: None,
                     rolled_back_patch_numbers: Some(vec![1]),
+
+                    available_release_versions: None,
                 })
             },
             UNEXPECTED_DOWNLOAD,
@@ -824,6 +832,7 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: None,
+            available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_TESTS_PATCH.write_to(dest),
@@ -846,6 +855,8 @@ mod test {
                     patch_available: false,
                     patch: None,
                     rolled_back_patch_numbers: Some(vec![1]),
+
+                    available_release_versions: None,
                 })
             },
             UNEXPECTED_DOWNLOAD,
@@ -914,6 +925,7 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: None,
+            available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_TESTS_PATCH.write_to(dest),
@@ -936,6 +948,8 @@ mod test {
                     patch_available: false,
                     patch: None,
                     rolled_back_patch_numbers: Some(vec![1]),
+
+                    available_release_versions: None,
                 })
             },
             UNEXPECTED_DOWNLOAD,
@@ -959,6 +973,8 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: Some(vec![]),
+
+                    available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_TESTS_PATCH.write_to(dest),
@@ -1011,6 +1027,7 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: None,
+            available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_PATCH_2_PATCH.write_to(dest),
@@ -1035,6 +1052,8 @@ mod test {
                         hash_signature: None,
                     }),
                     rolled_back_patch_numbers: Some(vec![2]),
+
+                    available_release_versions: None,
                 })
             },
             |_url, dest: &Path, _resume_from: u64| HELLO_TESTS_PATCH.write_to(dest),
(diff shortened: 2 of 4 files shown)


lukemmtt:
Closing preemptively to keep this from cluttering the weekly triage

PR #358, opened 2026-05-14
fix(inflate): preserve real upstream cause + add diagnostic byte counters

bdero:
## Summary

Two fixes that together make `inflate()` failures actionable in the wild:

**1. Error attribution.** Today, `inflate()` always reports `Decompression of patch failed` when the decompression thread's `join` returns `Err`. That's misleading when the real failure was on the patching side — bipatch errors → `fresh_r` gets dropped → decompression thread's next write gets `BrokenPipe` → join returns `Err` → we report decompression as the primary cause. The actual root cause (the patch-side error) gets shadowed and disappears.

This PR flips the priority: when both error, `patch_result` is reported as primary and `decompress_err` is folded into the context chain (so a genuine zstd corruption error doesn't disappear either). The `(Ok, Err)` case — decompression reports a trailing error after bipatch already finished — is now treated as success with an info log; the output was produced and will be hash-checked downstream anyway.

**2. Diagnostic counters.** Every error path now carries a one-line snapshot of every stream in the pipeline:

```
patch_file=66515b base_size=2838304b
decompressed_into_pipe=2853167b decompressed_out_of_pipe=2853167b
base_read=2785583b base_last_seek_pos=25040 base_seeks=307
output_written=2851760b
```

Field meanings:

| Field | What it means |
|---|---|
| `patch_file` | on-disk compressed patch size |
| `base_size` | base reader length (probed via `seek(End(0))` at start) |
| `decompressed_into_pipe` | bytes the zstd thread wrote to the pipe |
| `decompressed_out_of_pipe` | bytes bipatch consumed from the pipe |
| `base_read` | bytes bipatch read from the base reader |
| `base_last_seek_pos` | where the base reader last seeked to |
| `base_seeks` | how many times bipatch seeked the base |
| `output_written` | bytes copied to the output file |

Reading the failing counter directly tells you where in the pipeline things went wrong. For example, a field crash where `base_read=0` immediately implicates the base reader; `decompressed_into_pipe=8 output_written=0` implicates bipatch parsing right after the 8-byte header; a `base_size` that disagrees with the size the host extracted while building the patch implicates install-time munging.

Counters are tracked via `Arc<AtomicU64>` shared with the decompression thread, wrapped around each stream end with three small `CountingReader` / `CountingReadSeek` / `CountingWriter` adapter structs. Format contract is pinned by `inflate_diagnostics_include_all_counters` so future refactors can't drop a field silently.

A secondary fix: the bipatch-init failure path used to early-`?` and leak the decompression thread detached. It now drains the thread and folds any concurrent decompression error into the same error chain.

## Motivation

Investigating a long-standing iOS-standalone patch failure that's hard to reproduce in the wild. On-device log lines were just `Update failed: Decompression of patch failed` — which (as it turns out) was downstream noise. With these counters the next investigator can read the failing pipeline state directly from the device log instead of having to instrument and ship a debug build.

## Test plan

- [x] `cargo build` clean
- [x] `cargo test inflate` — 5/5 pass:
  - `inflate_fails_with_corrupt_zstd_data_reports_bipatch_init_failure` (replaces the old "decompression or patch init" test)
  - `inflate_when_both_streams_error_reports_patch_side_primary` (replaces `inflate_reports_decompression_error_as_primary` — asserts new attribution)
  - `inflate_diagnostics_include_all_counters` (new — pins format contract)
  - `inflate_rejects_invalid_magic`, validation tests
- [x] `cargo test` — 247/247 pass
- [x] Verified diagnostic line shape against a real downloaded iOS patch + local App.framework/App base

## Out of scope

This PR only fixes the error reporting in `inflate`. The actual iOS-standalone patch failure is a separate investigation — these diagnostics are the prerequisite to making that investigation tractable.

### DIFF
--- a/library/src/updater.rs
+++ b/library/src/updater.rs
@@ -890,7 +890,7 @@ where
         )
     })?;
 
-    shorebird_info!("Base file size: {} bytes", base_total_size);
+    shorebird_debug!("Base file size: {} bytes", base_total_size);
 
     // Atomic counters shared across the decompression thread and this thread.
     let decompressed_into_pipe = Arc::new(AtomicU64::new(0));
@@ -1013,11 +1013,11 @@ where
     // doesn't disappear entirely.
     match (patch_result, decompress_result) {
         (Ok(_), Ok(_)) => {
-            shorebird_info!(
-                "Patch successfully applied to {:?}. {}",
-                output_path,
-                diag
-            );
+            // Match the previous info-level chattiness on the happy path:
+            // a single short success line. Detailed counters land at debug
+            // level for development / opt-in production diagnostics.
+            shorebird_info!("Patch successfully applied to {:?}", output_path);
+            shorebird_debug!("Inflate diagnostics on success: {}", diag);
             Ok(())
         }
         (Err(patch_err), Ok(_)) => {
--- a/library/src/updater.rs
+++ b/library/src/updater.rs
@@ -1013,11 +1013,16 @@ where
     // doesn't disappear entirely.
     match (patch_result, decompress_result) {
         (Ok(_), Ok(_)) => {
-            // Match the previous info-level chattiness on the happy path:
-            // a single short success line. Detailed counters land at debug
-            // level for development / opt-in production diagnostics.
-            shorebird_info!("Patch successfully applied to {:?}", output_path);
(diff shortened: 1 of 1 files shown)


eseidel (approved):
It's maybe a little over the top, but seems reasonable.

PR #363, opened 2026-06-29
fix: populate current_patch_number from running_patch, not the boot breadcrumb

easymac:
## Summary

`current_patch_number` (added in #343) is read from `currently_booting_patch()` at both patch-check sites (`updater.rs:285`, `:420`). That's a transient boot breadcrumb — `record_boot_success()` clears it before any Dart-initiated patch check fires, so the field is `None` and `skip_serializing_if` drops it from the wire. Result: it reaches the server on a negligible fraction of requests fleet-wide.

Fix: read from `running_patch()` — set at `report_launch_start`, held all session, survives rollback. Same value the FFI symbol `shorebird_current_boot_patch_number` already uses.

```diff
-            state.currently_booting_patch().map(|p| p.number),
+            state.running_patch().map(|p| p.number),
```
(both sites)

## Notes
- `cargo check` passes; no boot-state-machine change.
- Follow-up: raise the console's `3.41.9` floor once this lands in a Flutter release.

### DIFF
--- a/library/src/updater.rs
+++ b/library/src/updater.rs
@@ -282,7 +282,7 @@ pub fn check_for_downloadable_update(channel: Option<&str>) -> anyhow::Result<bo
     let (client_id, current_patch_number) = with_state(|state| {
         Ok((
             state.client_id().to_string(),
-            state.currently_booting_patch().map(|p| p.number),
+            state.running_patch().map(|p| p.number),
         ))
     })?;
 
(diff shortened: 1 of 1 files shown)


library/src/updater.rs
```diff
@@ -417,7 +417,7 @@ fn update_internal(_: &UpdaterLockState, channel: Option<&str>) -> anyhow::Resul
         Ok(PatchCheckRequest::new(
             &config,
             &state.client_id(),
-            state.currently_booting_patch().map(|p| p.number),
+            state.running_patch().map(|p| p.number),
```
bdero, line 420: nit: can we unittest this?

bdero (approved):

library/src/updater.rs
easymac: Added a unit test 👍

bdero (approved):
lgtm

PR #364, opened 2026-07-02
feat(updater): classify update failures into stable categories

eseidel:
## What

Adds internal, typed classification of update failures in the Rust updater. Today every failure is flattened to `SHOREBIRD_UPDATE_ERROR` at the FFI boundary, which is why the Dart `UpdateFailureReason` collapses network / permission / disk-full / server failures all into `unknown`. This change recovers that information **below** the FFI boundary.

- New `FailureCategory` enum (`Network`, `Server`, `OutOfSpace`, `PermissionDenied`, `Filesystem`, `CorruptPatch`, `Unknown`) with a stable `code()` (telemetry key) and a `retryable()` mapping.
- New `classify_failure(&anyhow::Error)` that walks the error chain by downcast. **Unambiguous OS error kinds win first**, so a disk-full while inflating a patch reports `OutOfSpace`, not `CorruptPatch`.

To make classification reliable, error origins now **preserve typed causes** instead of flattening to strings:
- `network.rs`: `handle_network_result` returns a typed `NetworkError` (`Server { status }` vs `Transport`) rather than `bail!`-ing a string. **Messages are preserved verbatim.**
- `updater.rs`: adds `UpdateError::CorruptPatch`, tagging hash-mismatch, zstd-validation, and download-size-mismatch failures. Tags are attached as **inner markers** (via `.context`) so the detailed, developer-facing message stays the top-level `Display` — no user-visible or telemetry message text changes.

## The payoff (no C ABI change)

The category is consumed only to enrich the existing server-bound failure diagnostic event:

```
update_failure: patch 42 failed [corrupt_patch] (Downloaded patch was invalid: ...)
```

So we get categorized failure telemetry **immediately, with zero FFI/Dart/header changes** (verified: the diff touches only `network.rs` and `updater.rs`).

## Why not expose new FFI error codes

This is deliberately kept off the C ABI. The `shorebird_code_push` v3 PRD plans a customer-facing `Failed{category, code, retryable}` (Change #7), exposed via a **new structured-result C symbol**, and explicitly rejects transitional 2.x FFI shapes ("we go straight to 3.0") since every Rust C symbol is a permanent contract. This PR is the **enabling layer** for that surface: it proves the taxonomy and the `category → retryable` mapping without spending permanent-contract budget. FFI exposure is deferred to v3.

## Testing

- `cargo build`, `cargo clippy --all-targets`, `cargo fmt --check` — all clean.
- `cargo test` — 258 pass, 0 fail. Includes 8 new classifier tests (per-category, the OS-kind-priority ordering, and context-wrapping preservation).

## Notes / out of scope

- Pre-existing: `NetworkError::Transport`'s message still says "Patch check request…" even for download/event failures (TODO preserved). Easy to fix here if desired.

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -39,6 +39,7 @@ words:
   - dllexport
   - dlopen
   - dorny
+  - downcastable
   - downloadables
   - EACCES
   - EDQUOT
@@ -71,6 +72,7 @@ words:
   - pubspec
   - repr
   - reqwest
+  - retryability
   - rlib
   - roundtrips
   - rsplit
--- a/library/src/events.rs
+++ b/library/src/events.rs
@@ -80,6 +80,13 @@ pub struct PatchEvent {
     /// An optional message to be sent with the event.
     /// Care should be taken that this field *never* contain PII or sensitive information.
     pub message: Option<String>,
+
+    /// For failure events, a stable, low-cardinality classification of the
+    /// failure (the `FailureCategory` code, e.g. "network", "corrupt_patch").
+    /// Sent as structured data so the server does not have to parse it back out
+    /// of `message`. Optional and ignored by servers that don't yet read it.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub failure_category: Option<String>,
 }
 
 impl PatchEvent {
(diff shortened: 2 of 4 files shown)


eseidel:
This will get exposed as part of our v3 work for package:shorebird_code_push.  I started this today because a large enterprise customer was seeing (small numbers of) errors in the wild and wanted better explanations/categorization for them.

PR #373, opened 2026-09-11
fix: latch the running patch before the update thread starts

easymac:
**Problem**

The patch check that the update thread sends on launch omits `current_patch_number`. The running patch was recorded in `shorebird_report_launch_start`, but the engine calls `shorebird_start_update_thread` from `ConfigureShorebird` and does not report launch start until the Dart VM resolves the isolate snapshot. By then the first check has usually gone out, so the server sees a device on an unknown patch.

**Change**

`shorebird_next_boot_patch_path` now also records the returned patch as the one this process is running. The engine calls it once per `ConfigureShorebird`, immediately before starting the update thread, so the check carries the patch the engine is about to load. The first call in a process wins: the update thread can install a newer patch before the VM boots, and a later caller must not relabel the process as running it. `shorebird_report_launch_start` still latches, as a no-op when already set, for embedders that never ask for the path.

The Dart-facing `shorebird_next_boot_patch_number` keeps its non-latching path. App code can call it at any time, including after a newer patch installs, and routing it through the latch would break the current-versus-next comparison behind `restartRequired`.

No engine change is needed. `ConfigureShorebird` already validates, asks for the patch path, and then starts the update thread.

**Not in this PR**

The boot breadcrumb that `shorebird_report_launch_start` writes still comes from `next_boot_patch`, so in the same race it can name a patch this process never booted, and launch success or failure then lands on the wrong patch. That predates this change and needs the latch to exist before it can be fixed. A follow-up PR will switch the breadcrumb to the latched patch.

A process with more than one engine can have a second `ConfigureShorebird` hand a different patch path to the second engine. The latch reports the first. That is what a single per-process value can express and matches what `shorebird_current_boot_patch_number` has always meant.

**Testing**

Three tests in `patch_check_current_patch_number_tests`: the engine's call order sends the running patch on the wire, a patch installed during boot does not relabel the process (and pins the known-wrong breadcrumb for the follow-up), and a base-release boot stays latched as base when a patch installs during boot. `cargo test` passes in `library`, `cargo clippy --all-targets` is clean, and `updater_engine.h` is regenerated by the build.


eseidel:
Thanks for tracking this down, the race is real and the writeup is clear. Rather than teach Rust to latch the running patch before the update thread starts, I'd remove the race: there's no reason the update thread needs to start before the boot is recorded. It's a background check against our server that applies at the next launch, and it's happy to lose the race with boot every time.

shorebirdtech/flutter#164 moves the thread start out of `ConfigureShorebird` into the shim's once-guarded `ReportLaunchSuccess`. That makes the launch check carry the running patch with no Rust change, and it also closes the lifecycle hazard your "not in this PR" section touches on: an install completing during a first boot retires the patch being booted (the lifecycle only protects `last_booted`), which is why the breadcrumb could name a patch that never ran. After success that can't happen, so no breadcrumb change is needed either.

Same principle as moving `ReportLaunchStart` out of configure: configure is "an engine might boot", the launch reports are "an engine did boot", and lifecycle side effects should key off the latter. This PR is superseded by #164; would appreciate your review there. We're not backporting; this only affects Mission Control attribution, and I'll raise the console's Flutter floor to the release that carries the fix.

PR #166, opened 2024-05-20
feat: adding CI for updater tools

erickzanardo:
## Description

While getting up to speed on the `updater_tools`, I noticed it didn't had a workflow for itself. This PR adds one.

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #375, opened 2026-09-15
feat: start the update thread from the launch-success report

eseidel:
Alternative to shorebirdtech/flutter#166, for comparison. Rust half; the engine half is the paired shorebirdtech/flutter PR.

## Problem

The engine's C++ `Updater` shim keeps a shadow of the boot lifecycle in static booleans (`launch_started_`, `launch_completed_`, `initialized_`) and, since flutter#164, starts the update thread from behind the success guard. That guard is also claimed by a failure report, so a patch that fails to load (failure from `TryLoadFromPatch`, then base code boots and reports success) never checks for updates that launch. flutter#166 fixes it by adding a fourth boolean. This PR moves the decision to where the lifecycle already lives.

## Change

- `report_launch_success` records the boot if one was in progress, then calls `maybe_start_update_thread`, which starts the thread once per process when `auto_update` is on. An unconfigured updater leaves the guard alone so a later `init` in the same process still gets its check.
- `report_launch_failure` clears the running patch. The engine boots base code after a patch fails to load, and the check that launch sends must not name the failed patch. This is the `running_patch` change flutter#166's rollout note depends on.
- `shorebird_start_update_thread` and `shorebird_should_auto_update` stay exported. The engine stops calling them in the paired PR; removing them is a follow-up once no engine pins a revision that does.

## Launch sequence

Device has patch 3 booted, patch 4 installed, patch 5 live on the server. Patch 4 fails to load.

Before: launch 1 marks 4 bad, boots base, no check. Launch 2 boots 3, checks, fetches 5. Launch 3 boots 5.

After: launch 1 marks 4 bad, boots base, checks, fetches 5. Launch 2 boots 5.

## Rollout order

Land this first. An engine pinned here that still calls `shorebird_start_update_thread` itself would start two threads, so the engine PR removes that call and rolls `DEPS` in one commit.

## Testing

- 253/253 `cargo test -p updater --lib`; clippy clean with `-D warnings`.
- Under `cfg(test)` the spawn is counted rather than run, so the tests do not race a real update against the next test's config reset. The production spawn is the one-line call to the existing `start_update_thread`.
- Mutation-verified:
  - Guard hoisted above the configured check fails `launch_success_before_init_does_not_claim_update_thread_guard`
  - Thread start gated on a patch having booted (the flutter#164 shape) fails `launch_failure_then_success_starts_update_thread`
  - Guard removed fails `launch_success_starts_update_thread_once`
  - `running_patch` not cleared fails `report_launch_failure_with_patch`


eseidel:
Closing; shorebirdtech/flutter#166 lands instead. The `report_launch_failure` change that clears `running_patch` is still needed for #166's rollout and can be lifted from here into its own PR.

PR #376, opened 2026-09-15
fix: clear the running patch when a launch fails

nickshorebird:
## Summary

`report_launch_start` records the booting patch as the running one. A launch failure marks that patch bad and the engine falls back to base code, but `running_patch` kept pointing at it, so the patch check that launch sends named a patch the process was not executing. Nothing read the stale value before, because the update thread never ran on the failure path, and shorebirdtech/flutter#166 starts it there. I lifted these lines from #375 rather than rewriting them, since that PR is closed for reasons unrelated to this part.

- `report_launch_failure` clears `running_patch` after marking the patch bad

## Testing

- 249/249 `cargo test --lib`
- Removing the `set_running_patch(None)` line fails `report_launch_failure_with_patch`, so the assertion is load-bearing rather than decorative

## Rollout order

The engine only picks this up when `updater_rev` in the Flutter fork's `DEPS` moves to it. shorebirdtech/flutter#166 does not depend on this and can land first. Until both are in and `DEPS` is rolled, a device that falls back to base code reports the failed patch as running.


nickshorebird:
@eseidel - FYI force merging. Lines pulled from your PR which I agree with.

